PR #1, opened 2017-12-05
GraphQL Subscriptions FTW

nazar:
Implemented a basic `games_messages` table with model and schema and setup a client `/games/:id` route that shows specific games with a super basic message editor to get graphQL subscriptions working.

Open two windows for a specific `/games/:id`, type a message and watch the messages be updated in both windows via websocket. 🎉


api/app.js
```diff
@@ -53,14 +59,18 @@ if (config.get('graphql.graphiql')) {
     context: { req, res }
   })));
 
-  app.use('/graphiql', graphiqlExpress({ endpointURL: '/graphql' }));
+  app.use('/graphiql', graphiqlExpress({
+    endpointURL: '/graphql',
+    subscriptionsEndpoint: `ws://localhost:3001/subscriptions`
```
mikelax, line 64: @nazar  does this need to come from a config value?

mikelax:
in starting up the `api` container, I received this error after starting this branch. 
```
[3/4] Linking dependencies...
warning "graphql-redis-subscriptions > graphql-subscriptions@0.4.4" has incorrect peer dependency "graphql@^0.7.0 || ^0.8.0 || ^0.9.0 || ^0.10.1".
```
is this a dep in our end we can corret?

api/app.js
nazar: definetly

nazar:
looks like latest graphql-subscriptions is 0.5.5 - looks like it is being pulled in as a dep of another package - hopefully it will fix itself....

mikelax:
@nazar  I noticed that the created at and updated db columns are null for both games and games_messages tables. 

I think that we need to pass two `true` parameters to the `table.timestamps` [function](http://knexjs.org/#Schema-timestamps).

nazar:
@mikelax - re the timestamps - I'll do that in another branch. objective is pretty low level and doesn't populate these columns without a plugin. I'll add that probably in the next PR.

Same thing for the deleted_at column

nazar:
Oh I see it - nice for the `true, true` - looks like it add the timestamp to the table DDL as a default value - 👍

mikelax:
@nazar  yeah double `true` params was the trick to get them to have defaults and be not null. 
only question is does the updated_at get updated automatically or do we have to handle that in our sql.

nazar:
>only question is does the updated_at get updated automatically or do we have to handle that in our sql.

Good question - I'll have a play on my local and see what it looks like. We might need something like [this](https://github.com/Vincit/objection.js/tree/master/examples/plugin) if we need to.

nazar:
Hey @mikelax - can we merge this PR? There are some minor issues but these can be addressed individually in smaller tickets.

Issue #2, opened 2017-12-05
Implement react Form select on create/edit game screen


mikelax:
Instead of using textfield elements for the skill and posting frequency attributes, they should instead be [select elements](https://reactjs.org/docs/forms.html#the-select-tag). 

Options to use for Skill Level. The first option can be the default on initial load of the page

- Any/Newbie friendly, 
- I’ve rolled dice before
- Expert/role play master and rules bookworm

Options to use for Posting Frequency. NOTE: also need to update label as it currently says Frequence. 

- About 1 / day
- 2-3 times / week
- Hardcore - More than 1 / day

Let's also update the default (initial state) values for the min players to 2, and max players to 6.

PR #11, opened 2017-12-07
#2 - Implement react Form select on create/edit game screen

nazar:


client/src/components/Games/Create/CreateGame.js
```diff
@@ -24,97 +35,128 @@ const CreateGame = class CreateGame extends Component {
...
-  }
+        })
+        .then(() => this.setState({ saved: true }));
+    }
+  };
+
+  valid = () => {
+    const errors = {};
+
+    _.isEmpty(this.formValue('title')) && (errors.title = 'Title is required');
```
mikelax, line 132: @nazar won’t these errors overwrite each other if there is more than one?  Shouldn’t the form input key be included in the errors prop?

client/.eslintrc
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": "react-app"
+  "extends": "react-app",
+  "rules": {
+    "semi": 2     // come on! Gotta have ;
```
mikelax, line 4: How was this not included in their ruleset?? :thinking

client/src/components/Games/Create/CreateGame.js
```diff
@@ -24,97 +35,128 @@ const CreateGame = class CreateGame extends Component {
...
-              onChange={e => this.setState({ skillLevel: e.target.value })}
-            />
+              componentClass="select"
+              value={this.formValue('gameSettings.skillLevel')}
+              onChange={this.formInput('gameSettings.skillLevel')}>
+              <option value="1">Any/Newbie friendly</option>
+              <option value="2">I’ve rolled dice before</option>
+              <option value="3">Expert/role play master and rules bookworm</option>
+            </FormControl>
+
```
mikelax, line 91: Can you update this word to Frequency?

client/src/components/Games/Create/CreateGame.js
```diff
@@ -24,97 +35,128 @@ const CreateGame = class CreateGame extends Component {
...
             <FormControl
-              type="text"
-              value={this.state.postingFrequency}
-              placeholder="Posting Frequence"
-              onChange={e => this.setState({ postingFrequency: e.target.value })}
-            />
+              componentClass="select"
+              value={this.formValue('gameSettings.postingFrequency')}
+              onChange={this.formInput('gameSettings.postingFrequency')}>
+              <option value="1">About 1 / day</option>
```
mikelax, line 97: @nazar what happens if we have to display the selected value elsewhere in app?  

Or on the search results page?  Should they (display values) be contained in a common location

mikelax:
also forgot to ask ,  is the call to this.forceUpdate() an Apollo or react function?

Are you using that because you didn’t use setState function?

client/.eslintrc
nazar: I'm not surprised - this is what  happens when the java/c# class/deep nested inheritance trees crowd flock to javascript. First mixins are outlawed and everything turns into HoC of a HoC of a HoC (see a pattern?)

Joking! 😸

client/src/components/Games/Create/CreateGame.js
nazar: That's a simple refactor to extract the key and descriptions to a function which can be reused in other places in the app.

client/src/components/Games/Create/CreateGame.js
nazar: oops - typo

nazar:
>is the call to this.forceUpdate() an Apollo or react function?

`ForceUpdate` is a react function - these should no longer be present in the PR

mikelax (approved):
🚀

Issue #7, opened 2017-12-06
Implement Game/Forum Post Interface


mikelax:
Overview
===
This ticket covers the work to integrate the main "create post" text window. It will be implemented as a type of rich text editor (RTE). This will be one of the most common interactions the user will have with the site so the UX and how well it works will be critical. 

Requirements
===

- The editor should easily allow the user to create the different types of "text", and actions within a single post. 
- The initial set of text types are:  Character Talking, Character Thinking, Character Signing, Regular Text. 
- As discussed on a call, a game will primarily consist of a single game thread. This thread will be made of posts of different types. Everything will be displayed chronologically in a single feed so its easy to always have context around a post. 
   - The initial post types will be:  In Character (IC), Out of Character (OC)
   - TODO is how to handle additional types such as dice rolls, DM actions such as "call for initiative"
   - This will be handled in separate tickets, it's listed here just for context around this chat interface. 
- The best UX for this interface is that the text markup renders correctly as the user is creating the post. If this proves too difficult, the next best option would be to give them a "preview" tab that they can easily switch between typing and previewing. Example UX would be the github issue/comment UI widget

Possible libraries
---
[react-quill](https://github.com/zenoamaro/react-quill)
[draftjs](https://draftjs.org/)
[react-rte](https://github.com/sstur/react-rte) - based on draft js. Not sure if we need this or would write our own similar to this lib based on draft.  
[this reddit thread](https://www.google.com/amp/s/amp.reddit.com/r/reactjs/comments/5o6eh6/which_wysiwyg_editor_is_recommended_with_react/) talks about history of draft. Lots of projects after it was released but a lot are stale now. Need to research them more. Draft is the building blocks to create own editor. Careful though as there are mobile issues and shims possibly needed for create react app. 
[marked](https://github.com/chjj/marked)

mikelax:
@nazar  one thing I was thinking about this am, as you are going through the changes with draft, once you get to a logic point you should make sure that the full utf8 unicode char set is allowed. ie. we want to allow full emoji and special character usage. If we need to make any db conf changes to set the db in the correct charset or collation we should do so to make things easier.

nazar:
Yep yep re utf-8 and emojis. Draft supports emoji's provided the page has a `<meta charset="utf-8">` in `<head>`, which it has.

I _think_ pg `utf-8` support 4bit unicode - we'll just have to make sure the DB is set to `utf-8` charset when it is created.

[...]

PR #18, opened 2017-12-13
Refs #7 - Implement Game/Forum Post Interface

nazar:
PR contains DraftJS GameMessage editor, with specific sample stylings for `shout`, `think` and `sing` - these samples are for demonstration purposes. The default status is `talk`

Only in-line styles have been implemented but block level styles (i.e. paragraph based elements such as headers) are to be implemented. One such example could be a `singing` block level element that takes a selection of several lines and turns it into a paragraph, with italics and centered text.... I'm sure we can think of more block level elements.

In character/ooc not implemented for this PR - we will revisit once character banners are in place (or before)


mikelax:
Awesome @nazar. Can’t want to look at this.  
One comment though with regards to the styling. 

Even for the In character posting, I was thinking the default mode or text format would be that of “narrator”. 

Ie descriving what the character may be doing. If you look at my posts on BGG. You’ll see i narrate movements, actions, etc and then mix it up with character talking, thinking, etc.  

So I think the default node would be narrate and then  character speaking is just another option like singing, etc.

nazar:
Yep - setting default/initial styles won't be an issue. We should be able to heavily customize the editor.

api/migrations/20171209140822_change_messages_message_to_json.js
```diff
@@ -0,0 +1,19 @@
+exports.up = function(knex) {
+  return knex('game_messages')
+    .del()
```
mikelax, line 3: scary!!, ha don't want to do this later on in the project dev.

client/src/components/Games/View/GameMessages.js
```diff
@@ -44,7 +45,7 @@ class GameMessages extends Component {
     return <div className='game-messages'>
       {_.each(gameMessages).map(({ id, message }) => (
         <div key={id} className='game-message'>
-          {message}
+          <GameMessage message={message} readOnly={true} />
```
mikelax, line 48: 👍

client/src/components/Games/components/GameMessage/GameMessage.js
```diff
@@ -0,0 +1,258 @@
...
+}
+
+function onToggleAction(entityKey) {
+  const editorState = this.state.editorState;
+  const currentContent = editorState.getCurrentContent();
+  const selection = editorState.getSelection();
+  const selectionEntity = getSelectionEntity(editorState);
+  const collapsed = selection.isCollapsed();
+
+  // many many many options here
```
mikelax, line 94: 😨

client/src/components/Games/components/GameMessage/GameMessage.js
```diff
@@ -0,0 +1,258 @@
...
+    e.preventDefault();
+    this.props.onToggle(this.props.entity);
+  };
+}
+
+/*
+define individual styles as functional components
+*/
+
+const ThinkSpan = (props) => {
```
mikelax, line 186: @nazar  do you want to document here or maybe in a readme section what is involved in adding a new style, ie. all the files that need to be updated?

mikelax (commented):
@nazar  this is really great!
amazing work on getting this message component functional. i think maybe in a separate ticket we can work on the actual buttons and styles we want to launch with.,

client/src/components/Games/components/GameMessage/GameMessage.js
nazar: not just yet @mikelax as this component is likely change significantly - we'll definitely add documentation once we have a better idea of what components we'll need 📖

mikelax: @nazar gotcha, makes sense. I left you a couple of messages on slack if you are around. i can help look at these two PRs and we can try to get them both merged.

mikelax (approved):
👍

Issue #3, opened 2017-12-05
Implement shadow (local) users table in postgres


mikelax:
## Overview

After doing some reading on various tutorials and [posts](https://community.auth0.com/questions/802/few-questions-regarding-crud-and-users), I think that we will need to implement a users table in the main postgres database. This table will store the Auth0 user_id column, along with a set of metadata needed by the website to function and provide all data to the UI for display (ie. user profile image). 

We will implement a similar solution as the community answer above that creates a two-way link between the pg users table and Auth0 users table. 

## Development Tasks

- [x] Migration for pg users table & foreign keys
- [x] Auth0 rule for login
- [x] (Rest) API to create user record in pg table. Params of Auth0 user_id, user and app metadata, response of pg user id
- [x] Add users seed and refactor games seed
- [x] Refactor games and messages insert to include user id

## Questions:

1. Do we need to create a user id (pk) attribute in pg table, or can we use the Auth0 user_id field as the primary key in the pg users table?  Yes, we will use internal pg userId. It will be added to Auth0 app_metadata

I created [this question on the Auth0 community](https://community.auth0.com/questions/12656/best-practice-for-storing-user-profile-attributes) site to see if I can get any better answers that what I was able to find from random googling.

nazar:
>Do we need to create a user id (pk) attribute in pg table, or can we use the Auth0 user_id field as the primary key in the pg users table?

That depends on how auth0 can handle a single user that uses multiple auth methods - i.e. all of google, twitch and email. It also depends on whether auth0 provides a mechanism that allows a user to merge say google and twitch logins into a single profile and always return the same user_id despite the user's choice of auth provider.

In terms of `users.id` I think all tables should have a pk based on a sequential INT id - if anything just for consistency that all tables have an INT id as that would simplify the underlying code. If auth0 allows for the both the scenarios covered in the prev paragraph then we can record the auth0 user id and use that as an alternate key. We might need an additional table if auth0 does not cover the scenarios described above - I haven't had a chance to deep-dive into auth0 docs yet :)

mikelax:
@nazar the good thing is that auth0 does autmatically handle all the use cases you pointed out. 
Based on the OIDC implementation, we can guarantee that a user_id we receive back from auth0 will be unique. This includes all different auth methods we choose to support, if a user auths with one mechanism, then "adds" another method. 

[...]

PR #19, opened 2017-12-16
#3 WIP Add local users table. Need to secure API.

mikelax:
Addresses #3 
Just need to secure the POST /users API. Can't figure out for sure if we can get an access token within the scope of a Rule, so maybe just have to send shared secret in payload to be validated with custom middleware or inside route itself?


mikelax:
@nazar can you show me the example in this branch of how to update the Games (or messages) with the userId from the access token. It's the `sub` attribute in the token.
After the express-jwt middleware I think it would be accessible with `req.user.sub`.

nazar:
Hey @mikelax ,

The `context` object [here](https://github.com/mikelax/forgingadventures/blob/master/api/schemas/game.js#L57) (and other graphql resolvers) contains both the `req` and `res` Express objects. The token info should be available here on the `req` object.

In Match we have a [middleware](https://github.com/WhoSay/dataslinger/blob/master/app/middleware/getUser.js) that detects the token and fetches the user object from the database and adds it to the `req` object, from where we access it further down in the route implementation code.

HTH

nazar (approved):
:120%:

mikelax:
@nazar  I updated this PR to add the `userId` attribute to the `input` graphql object. 
I made a service instead of middleware as "for now" I didn't want to have to do a db hit for a user lookup on *every* graphql request when the majority of them will most likely be reads.

I think that we might be able to create a new function to do the user lookup and amend the `input` all in a single function if we want to refactor later.

PR #20, opened 2017-12-23
Issue 5 #5 add user to roles after sign up.

### LINKED ISSUE #5: Assign new user after initial sign up to correct Auth0 role
## Overview

After a user completes the sign up process, they must be automatically added to the correct Role within Auth0. The process should be automated for general users that sign up for the site. Granting users the administer Role can be a manual task done through the Auth0 Dashboard as it will be an infrequent change. 

## Requirements

Document in the Auth0 README the process for making a user an "admin". ie. granting the Auth0 Administers role with the [authorization extension](https://forgingadventures.us.webtask.io/adf6e2f2b84784b57522e3b19dfc9201/roles).

mikelax:
New Rule to add User to correct "User" Role after initial sign up. 
Fixes #5

### DIFF
--- a/auth0/README.md
+++ b/auth0/README.md
@@ -30,9 +30,25 @@ Github issue #10 covers the work to integrate automated deployments of Rules bas
 
 API_BASE_URL - for local development you can use [localtunnel](https://localtunnel.github.io/www/) to expose your localhost API to Auth0. 
 
+SHARED_SECRET - symetric shared key to use for calling FA API. It should match the auth0.sharedSecret config attribute.
+
+AUTH_EXT_BASE_URL - The API URL for the Auth0 [authorization extension API](https://auth0.com/docs/api/authorization-extension?http#find-your-extension-url) (US-Wes).
+
+AUTH_EXT_TOKEN_URL - The API URL to get an API Access Token. ie. tenant name /oauth/token
+
+AUTH_EXT_CLIENT_ID - The "Auth Ext" Client ID
+
+AUTH_EXT_CLIENT_SECRET - The "Auth Ext" Client Secret
+
 ## auth0-authorization-extension
 
 This Rule is automatically created by enabling and configuring the Auth0 Authorization extension. The basic function is to determine what "extra" attributes will be added to the User context. Currently it is set up to include the `roles` and `permissions` defined for the given User.
+
+## Configure User After Sign up
+
+This Rule will add the newly signed up User to the correct "User" Role so they have the necessary Permissions once they are redirected back to the application. 
+
+This Rule only performs logic after a Sign up, after a regular login it is a no-op. It makes use of the [authorization extension api](https://auth0.com/docs/api/authorization-extension?http#find-your-extension-url).
 
 ## Validate Access Token Scopes
 
--- a/auth0/rules/configure-user-after-sign-up.js
+++ b/auth0/rules/configure-user-after-sign-up.js
@@ -1,0 +1,88 @@
+function (user, context, callback) {
+  var _ = require('lodash');
+ 
+  // US West Region
+  var AUTH_EXT_BASE_URL = configuration.AUTH_EXT_BASE_URL;
+  var AUTH_EXT_TOKEN_URL = configuration.AUTH_EXT_TOKEN_URL;  // tenanturl/oauth/token
+  
+  if (user.roles && user.roles.length) {
+    // user already configured, short-circuit out
+    return callback(null, user, context);
+  }
+
+  // Get access token for Auth Ext API
+  request.post({
+    url: AUTH_EXT_TOKEN_URL,
+    headers: {
+      'content-type': 'application/json' 
+    },
+    json: {
+      client_id: configuration.AUTH_EXT_CLIENT_ID,
+      client_secret: configuration.AUTH_EXT_CLIENT_SECRET,
+      audience: 'urn:auth0-authz-api',
+      grant_type: 'client_credentials'
+    }
+  },
+  function(err, response, body) {
+    if (err) return callback(err);
+
+    var accessToken = body.access_token;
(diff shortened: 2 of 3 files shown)


mikelax:
@nazar  I tried to promisify the request object but the Rule was timing out (callback not being called)?. Things look ok to me though. 😕 

**Rule testing Bluebird promises**
```js
function (user, context, callback) {
  var Bluebird = require('bluebird');
  
  // Bluebird.promisifyAll(Object.getPrototypeOf(request), {multiArgs: true});
  // var request = Bluebird.promisifyAll(require("request"), {multiArgs: true});
  Bluebird.promisifyAll(request, {multiArgs: true});
  
  var AUTH_EXT_BASE_URL = configuration.AUTH_EXT_BASE_URL;
  var AUTH_EXT_TOKEN_URL = configuration.AUTH_EXT_TOKEN_URL;  // tenanturl/oauth/token
  
  if (user.roles && user.roles.length) {
    // user already configured, short-circuit out
    return callback(null, user, context);
  }

  console.log('Before first request');
  // Get access token for Auth Ext API
  request.postAsync({
    url: AUTH_EXT_TOKEN_URL,
    headers: {
      'content-type': 'application/json' 
    },
    json: {
      client_id: configuration.AUTH_EXT_CLIENT_ID,
      client_secret: configuration.AUTH_EXT_CLIENT_SECRET,
      audience: 'urn:auth0-authz-api',
      grant_type: 'client_credentials'
    }
  })
  .then(function(response, body) {
    console.log('Before second request');
    var accessToken = body.access_token;
    // Get List of Roles
    return request.getAsync({
      url: AUTH_EXT_BASE_URL + '/roles',
      headers: {
        'content-type': 'application/json',
        'Authorization': 'Bearer ' + accessToken
      },
      json: true
    });
  })
  .then(function(response, body) {
    console.log('Num of roles is: ' + body.total);
    return callback(null, user, context);
  });
}
```

nazar:
Hey @mikelax 

>I tried to promisify the request object but the Rule was timing out (callback not being called)?. Things look ok to me though.

Is this code that runs on Auth0's infrastructure? If so, then `require` might not be available to us there. Either that or `request` needs to rebind `this` when promisified as not all method callbacks are promiseable - i.e. an example [here](https://github.com/WhoSay/whosay-stream/blob/master/lambda/whosay/lib/kafka.js#L27).

Worst case, try wrapping the function function in a `new Promise((resolve, reject) => {})` block, like [here](https://github.com/WhoSay/whosay-stream/blob/master/lambda/whosay/lib/kafka.js#L15-L17).

mikelax:
ok, I'll give that another try. 

I did try a couple of different calls to the promisfy process. 
I think the request object is just made available. The rules run in a sandbox process where there is a [limited number of packages](https://auth0.com/docs/appliance/modules) already included. 

```js
Bluebird.promisifyAll(Object.getPrototypeOf(request), {multiArgs: true});
var request = Bluebird.promisifyAll(require("request"), {multiArgs: true});
```

```js
Bluebird.promisifyAll(request, {multiArgs: true});
```

mikelax:
😢 , tried a few different options but still just getting timeouts from callback not being called. 

this was the latest I tried, both with and without the `.bind(request)`, neither worked. 

```js
function (user, context, callback) {
  var Bluebird = require('bluebird');
  
  // Bluebird.promisifyAll(Object.getPrototypeOf(request), {multiArgs: true});
  // var request = Bluebird.promisifyAll(require("request"), {multiArgs: true});
//   Bluebird.promisifyAll(request, {multiArgs: true});
  
  var AUTH_EXT_BASE_URL = configuration.AUTH_EXT_BASE_URL;
  var AUTH_EXT_TOKEN_URL = configuration.AUTH_EXT_TOKEN_URL;  // tenanturl/oauth/token
  
  if (user.roles && user.roles.length) {
    // user already configured, short-circuit out
    return callback(null, user, context);
  }

  console.log('Before first request');
  // Get access token for Auth Ext API
  Bluebird.promisify(request.post.bind(request), {multiArgs: true})
    ({
//  request.postAsync({
    url: AUTH_EXT_TOKEN_URL,
    headers: {
      'content-type': 'application/json' 
    },
    json: {
      client_id: configuration.AUTH_EXT_CLIENT_ID,
      client_secret: configuration.AUTH_EXT_CLIENT_SECRET,
      audience: 'urn:auth0-authz-api',
      grant_type: 'client_credentials'
    }
  })
  .then(function(response, body) {
    console.log('Before second request');
    var accessToken = body.access_token;
    // Get List of Roles
    return Bluebird.promisify(request.get.bind(request), {multiArgs: true})
      ({
      url: AUTH_EXT_BASE_URL + '/roles',
      headers: {
        'content-type': 'application/json',
        'Authorization': 'Bearer ' + accessToken
      },
      json: true
    });
  })
  .then(function(response, body) {
    console.log('Num of roles is: ' + body.total);
    return callback(null, user, context);
  });
}
```

mikelax:
@nazar  think we should give up on the promise support in rules for time being? 
I think at some point we might migrate to "Hooks", which are the newer version of Rules and those have a lot more capability than "Rules" and the Rules sandbox.

nazar:
Hmmm... I'm not seeing the Bluebird library in the [provided modules](https://auth0.com/docs/appliance/modules) T_T. The closest thing to a promise library I see is [q](https://github.com/kriskowal/q), which is a very dated promise library - weird they picked that over Bluebird...

We either have to use callbacks or maybe the native ES6 Promise and `promisify = require('util').promisify`

mikelax:
@nazar  given that this is a small rule "snippet" in a sandbox that we don't have control over, I am inclined to just leave it as-is with callbacks. Once their "hooks" come further along we can migrate to those which I believe add promise support.

Issue #13, opened 2017-12-07
Basic client styling using css3 grids


nazar:
- [x] Add stylus with [rapture](https://github.com/jescalan/rupture) for media query sugar
- [x] determine app wide grid levels taking into account react-router 4 nesting
- [x] create game grid
- [x] list games grid
- [x] view game grid

nazar:
As discussed with @mikelax - there is a temptation to eject the application in order to utilise webpack to:

1. implement webpack stylus loaders. use [rapture](https://github.com/jescalan/rupture)
2. SSR - for v2
3. nice to have - [HMR](https://medium.com/@rajaraodv/webpacks-hmr-react-hot-loader-the-missing-manual-232336dc0d96)
4. add any additional webpack/babel functionality post eject

Concerns that need to be investigated before ejecting:

1. maintenance - what do we lose by not using RCA after ejecting
2. what does the code look like when ejected - i.e. when ejecting, does the app run exactly like it used to or is extra work needed to manually wire up all the magic from RCA

nazar:
![eject](https://cdn-images-1.medium.com/max/1500/1*vjt0c8jNe26ZmaIysFv3IA.jpeg)

nazar:
![css3 responsive grids](http://g.recordit.co/Kuaiw3UGQN.gif)

PR #24, opened 2017-12-28
#13 - set app wide container plus styling

nazar:
Hey @mikelax,

This PR contains an ejected app to support adding stylus plus stylus plugins as detailed in #13.

Although a few files were "ejected" the [stylus config](https://github.com/mikelax/forgingadventures/blob/25a1e3b56b7dbdc77ab57aa4134f3821d4839573/client/config/webpack.config.dev.js#L192-L238) is an exact copy of the [CSS config](https://github.com/mikelax/forgingadventures/blob/25a1e3b56b7dbdc77ab57aa4134f3821d4839573/client/config/webpack.config.dev.js#L159-L190) but with the addition of stylus loader options.

Going forward, I believe it will be straightforward to manually update CRA specific files as our configs are so far a simple addition to the existing configs.

I think ejecting the app is the way to go as it gives us more flexibility in setting/extgending our own configs.


client/src/components/Header/Header.js
```diff
@@ -1,82 +1,74 @@
...
-  }
+  };
 
   logout = () => {
     this.props.auth.logout();
-  }
+  };
 
   render() {
-    const { isAuthenticated, userHasScopes } = this.props.auth;
```
mikelax, line 16: should we set an eslint rule for this or did you fix this because you linted this code?

mikelax (approved):
💥 🚀 
@nazar  this is looking pretty good. do you think you can add a bit of documentation about some things we'll have to pay attention to, ie. lib upgrades, any webpack gotchas or config issues to remind us in three months?

client/src/components/Header/Header.js
nazar: Hey @mikelax - I suspect this was auto-formatted by my IDE - I'll double check my settings later today as I prefer spaces around braces

nazar:
>do you think you can add a bit of documentation about some things we'll have to pay attention to

👍

client/src/components/Header/Header.js
mikelax: @nazar  👍 
I am looking to run the app, specifically the api server, in the `test` env as I am getting close to having the CFN set up complete. I assume I need to set the `NODE_ENV` env var to `test`. 

Do I do that as a command before running the `npm start` command?

nazar: Hey @mikelax - `NODE_ENV=test npm start` should start you up in the `test` environment. Do note that the existing webpack config is development based. I'll need to add a test webpack configuration with slightly different code generation params. This test configuration will closely mirror the production configuration.

mikelax: 👍 I knew the pro webpack config would need to be updated, I'll write up a ticket to address the test/prod webpack configs. This will at least get me a server started to see if I can get the api server started in AWS env using *all* CFN resources (ie redis and RDS).

nazar:
@mikelax - can I go ahead and merge this PR?

mikelax:
@nazar I feel like we need a really big meme for this merge.   Eject!!  

Yeah go ahead and merge.  It’s looking great and sets up a nice base for app.

mikelax:
eject

Issue #32, opened 2018-01-07
Allow uploading user profile image


mikelax:
Overview
===
As a user, I want to be able to upload a different user profile image for my profile. 
NOTE: this process will also be very similar and used to allow the User to upload a "profile image" for each Character.

Requirements
===
* The user will be able to select and upload a new profile image on their account profile page.
* The image will be uploaded to an express API. This API will then: 
  * Upload file to S3
  * Upload file to Cloudinary
  * Make API call to Auth0 to set [picture attribute](https://auth0.com/docs/user-profile/user-picture
) in user_metadata.
* Web Client will make API call using `auth0-js` to Auth0 to refresh the local id_token which will now contain the new user profile image

For this ticket we will set up [environment variables in the docker compose](https://docs.docker.com/compose/environment-variables/#setting-environment-variables-in-containers) file to get api container access to S3. We will define a Role that will be used for dev as well as the higher environments. This Role will then be assigned to User objects. 

A separate ticket will be responsible for setting up and managing the AWS Keys needed to call S3 from the `api` docker container.

Tasks
===
- [x] Create Cloudinary Account
- [x] Create S3 buckets for each environment (or with subfolders) for image storage
- [x] Express API to `POST` profile image
- [x] Create [Auth0 Rule](https://auth0.com/docs/user-profile/user-picture#change-a-user-s-picture) to normalize `picture` attribute in tokens

mikelax:
Setting the NODE_ENV in the *api* project will be completed in #25

PR #33, opened 2018-01-08
#32 WIP for profile image upload.

mikelax:
Mostly working, here is a list of things left to implement: 

- [x] generate random filename so its the same for upload to S3 and cloudinary for consistency
- [x] set NODE_ENV=development when starting the api server
- [x] Figure out correct URL route for API, should it contain the db `id` value or Auth0 id?
- [x] Create files table in db, save record to it
- [x] Make API call to Auth0 to update user_metadata attribute
- [x] Auth0 Rule

@nazar  how did I do on the promises in `userImpl` functions?

### DIFF
--- a/api/config/development.json
+++ b/api/config/development.json
@@ -3,6 +3,11 @@
     "redirectUri": "http://localhost:3000",
     "sharedSecret": "SecretAdventuresInTheDungeon"
   },
+  "aws": {
+    "s3": {
+      "userUploadsPrefex": "development"
+    }
+  },
   "database": {
     "connection": "postgresql://postgres:docker@db:5432/postgres"
   },
--- a/api/config/development.json
+++ b/api/config/development.json
@@ -1,7 +1,9 @@
 {
   "auth0": {
-    "redirectUri": "http://localhost:3000",
-    "sharedSecret": "SecretAdventuresInTheDungeon"
+    "faClient": {
+      "redirectUri": "http://localhost:3000",
+      "sharedSecret": "SecretAdventuresInTheDungeon"
+    }
   },
   "aws": {
     "s3": {
--- a/auth0/rules/update-profile-with-custom-profile-image.js
+++ b/auth0/rules/update-profile-with-custom-profile-image.js
@@ -1,5 +1,5 @@
 function (user, context, callback) {
-  if (user.user_metadata.profileImage.imageUrl) {
+  if (user.user_metadata && user.user_metadata.profileImage && user.user_metadata.profileImage.imageUrl) {
     user.picture = user.user_metadata.profileImage.imageUrl;
   }
   
--- a/auth0/rules/update-profile-with-custom-profile-image.js
+++ b/auth0/rules/update-profile-with-custom-profile-image.js
@@ -1,0 +1,7 @@
+function (user, context, callback) {
+  if (user.user_metadata.profileImage.imageUrl) {
+    user.picture = user.user_metadata.profileImage.imageUrl;
+  }
+  
+  callback(null, user, context);
+}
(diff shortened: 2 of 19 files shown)


api/config/development.json
```diff
@@ -3,6 +3,11 @@
     "redirectUri": "http://localhost:3000",
     "sharedSecret": "SecretAdventuresInTheDungeon"
   },
+  "aws": {
+    "s3": {
+      "userUploadsPrefex": "development"
```
nazar: <3

mikelax:
@nazar  what do you think about the "user" id value to use in any users API endpoints. 
Ie. I would think this endpoint should be /users/:id/profile/image

Should the :id be the db id value or the Auth0 user id string? 
Probably should use the db id, but wondering if we have that value accessible on the client? 
it should be contained in the user_metadata attribute of the id_token, I *think*

nazar:
>Should the :id be the db id value or the Auth0 user id string?
Probably should use the db id, but wondering if we have that value accessible on the client?
it should be contained in the user_metadata attribute of the id_token, I think

I think it boils down to how we reference the `user` FK. For example, in our `games_players` table (not final name) that lists "accepted" users in a game will have something like `gameId` and a `userId` columns. The `userId` can either be sequence based FK or a String representing the auth0 `token_id`. 

Thinking about it in those terms, it would make more sense using `users.id` INT sequence based values as FKs tend to be sequential INTs. 

In regards to the `/users/:id/profile/image` endpoint: how will that endpoint be used by the API? Will it be called for every image we want to display for each user? If so, that would not be the best use of the endpoint as that data is better queried using graphQL using a query that looks something like:

```
    game(id: $id) {
      acceptedPlayers {
        user {
          name
          profileImageUrl
          ...
        }
      }
    }
```

which will return a list in a single trip.

mikelax:
@nazar yes you are right thatvat the db level we will use the db “id” column as FK for tables. Given that we should use the db “id” for the api url. 

This route will really only support POST to upload a new image. 
It will call Auth0 to set the metadata there for the user.  
Then we’ll be able to query via graphql like normal. 

I don’t  think we’ll need a separate column in the users table for this, it can read out of the user_metadata json column, right?

nazar:
>This route will really only support POST to upload a new image.

I always struggle deciding if `POST` endpoints should contain `:id` elements. Generally I don't but there are valid exceptions. Would a `PUT` be more appropriate as technically we are modifying an existing resource?

Also, I think for now it is easier (for an MVP) to just use the auth0 user_id for this - otherwise this will be blocked until we add a mechanism to query and store the user in the UI. We can always revisit this.

>I don’t think we’ll need a separate column in the users table for this, it can read out of the user_metadata json column, right?

Oh totally - that was just an example; sorry for the confusion.

nazar:
>@nazar how did I do on the promises in userImpl functions?

That's some serious Promise-fu right there Sensei!

6th Dan Black Belt

mikelax:
@nazar  this is functionally complete. 
Please send over any comments for refactor that might make sense as part of MVP. 
There are obv. lots of perf improvements that could be added. 

Also, it "feels" like a lot of code for the task of profile image upload, but I guess there are a lot of steps involved.

auth0/rules/update-profile-with-custom-profile-image.js
```diff
@@ -0,0 +1,7 @@
+function (user, context, callback) {
+  if (user.user_metadata.profileImage.imageUrl) {
+    user.picture = user.user_metadata.profileImage.imageUrl;
```
nazar: It might be safer to use the following form:

```
if (user.user_metadata && user.user_metadata.profileImage && user.user_metadata.profileImage.imageUrl) {
  ...
}
```

Although auth0 [provides lodash](https://auth0.com/docs/appliance/modules), unfortunately it is an older version that does not contain the `.get` function - i.e. `_.get(user, 'user_metadata.profileImage.imageUrl')`

nazar (commented):
Looks GREAT Mike!

I only left one tiny comment

nazar:
>Also, it "feels" like a lot of code for the task of profile image upload, but I guess there are a lot of steps involved.

I suspect the image upload would have involved more code if we tried to do it ourselves 👍

mikelax:
Closes #32

Issue #12, opened 2017-12-07
Tidy up client routes and auth object


nazar:
Tasks

- [x] refactor API JWT handling. Currently the [graphql](https://github.com/mikelax/forgingadventures/blob/master/api/middleware/security.js#L33) related middleware makes a call to `checkJwt` for each request, which adds a considerable amount of latency (between 200 and 800ms on my dev). This is either due to the library making a remote call or some other reason. In either case, the latency issue needs to be addressed via either correct use of the JWT auth0 library or via locally caching the JWT until it expires.

- [x] refactor [Auth](https://github.com/mikelax/forgingadventures/blob/master/client/src/components/App/App.js#L21) so that it can be passed as a prop to any component irrespective of its position in the DOM tree. 

- [x] Incorporate the auth state into the router so that login redirects can be implemented outside of components and views.

nazar:
> In either case, the latency issue needs to be addressed via either correct use of the JWT auth0 library or via locally caching the JWT until it expires.

Possibly related/solutions

* https://github.com/auth0/node-jwks-rsa#caching

mikelax:
@nazar it sounds like we do need to turn on caching.  
Part of verifying the checkJwt function involves making an api call to Auth0 to get a private key to decrypt the jwt.  

It sounds like caching in that lib is the way to go.

nazar:
Yep - I was using the jwt library incorrectly. It actually has remote key caching built-in and enabled by default. The way I was using it meant that the cache was re-built on each call #fail. Tis fixed now.

PR #34, opened 2018-01-08
Store the Auth state in redux and inject auth status into components

nazar:
Refs #12 - various

1. Add Redux to the client to manage the auth0 authorisation state and be able to inject it into any component
2. Update checkJwt API middleware to passthrough any requests that do not contain an `authorisation` header
3. Split the client Auth service class into two modules: `login.js` to manage auth0 Lock and `webAuth` to manage auth0-js
4. tidy client component folder by renaming each component's entry file to `index.js` - this allows us to reference the folder when using imports - i.e. `import Callback from '../Callback'` vs `import Callback from '../Callback/Callback'`
5. removed CRA artifacts such as the react logo etc...


client/src/actions/auth.js
```diff
@@ -0,0 +1,26 @@
+import {authLogout} from '../services/login';
+
+export const AUTH_FAILURE = 'auth_failure';
```
mikelax, line 3: @nazar  should these const values be snake case or camelCase?
ie. `export const AUTH_FAILURE = 'authFailure';`

client/src/components/Callback/index.js
```diff
@@ -0,0 +1,23 @@
...
+import { Redirect } from 'react-router-dom';
+
+function Callback(props) {
+  const {isAuthenticated} = props.authorisation;
+
+  if (isAuthenticated) {
+    return <Redirect to="/"/>;
+  } else {
+    return <div className="text-center">
+      Authorising
```
mikelax, line 12: s -> z

haha, its all over. not a big deal until we get newer design updates.

client/src/components/Messages/Create/TestPing.js
```diff
@@ -1,41 +0,0 @@
...
-export default class TestPing extends Component<{||}> {
-  constructor(props) {
-    super(props);
-
-    this.state = {};
-  }
-
-  securedPing = () => {
-    const { getAccessToken } = this.props.auth;
-    const API_URL = 'http://localhost:3000/api';
```
mikelax, line 15: @nazar  obviously this was always test code just for proof of concept in calling the api. 
Do you have an example of where the client is making an API call using "correct" pattern of constants for the URL etc?

mikelax (commented):
@nazar one additional question related to `AuthGuard.js`
I see how that is a HOC that wraps a given component that needs an Auth otherwise it redirects to /login

I didn't see though where the specific check for scopes are done for each component? ie. create game will use different scope than create character or List Games for example?

mikelax:
looking good @nazar . this is really nice! 
can you add a bit more info to the docs about some of the functions that are "redux" related I think. 
Things like the `mapStateToProps` or other redux related idioms?

client/src/services/webAuth.js
```diff
@@ -45,6 +45,17 @@ export function clearRenewalTimer() {
...
 }
 
+export function userHasScopes(scopes) {
+  if (!localStorage.getItem('scopes')) {
+    return false;
+  }
+
+  const grantedScopes = JSON.parse(localStorage.getItem('scopes')).split(' ');
+
+  return scopes.every(scope => grantedScopes.includes(scope));
```
mikelax, line 55: @nazar  is it correct to use the the `every` function here? Isn't the logic that the user must have at least one of the required scopes? Isn't that what we did for DS?

As an aside is this function `userHasScopes` referenced anywhere? I did a search of code and nothing came up?

client/src/actions/auth.js
nazar: It doesn't matter what the string value is as it is never directly accessed - only the constant is.

client/src/services/webAuth.js
nazar: `userHasScopes` isn't referenced anywhere; I copied it from the Auth class for "safe keeping" until we need it :)

client/src/components/Callback/index.js
nazar: hehe - I was planning on adding a big spinner (FontAwesome spinner came to mind) - but I didn't want to pull in a font icon library just yet

nazar:
>I didn't see though where the specific check for scopes are done for each component? ie. create game will use different scope than create character or List Games for example?

Yep - I haven't added scope checks yet

mikelax:
@nazar  thanks for the README info on redux, thunk, etc. very helpful. 🚀

mikelax (approved):

nazar:
oh 🤦‍♂️ - I just realised there is a typo in `authorisation` -> `authorization` 🤦‍♂️🤦‍♂️🤦‍♂️

Issue #25, opened 2017-12-29
Refactor webpack configs for test/prod environments


mikelax:
Overview
===
Now that we have ejected from CRA, we'll need to make some tweaks to the `client` webpack config. We *will* also need to make some updates to the `api` webpack config as well. 

Requirements
===
Update configs, @nazar  to provide more info if required. 

Client
---
TODO

API
---
Update the start script to set the NODE_ENV environment variable. When running in dev mode it should be set to "development". 

I *think* that we will be able to set environment variables such as this when you run a Task (container).

PR #35, opened 2018-01-09
refs #25 - Refactor webpack configs for test/prod environments

nazar:


nazar:
Hey @mikelax,

Re the naming of the webpack env based files (i.e. `webpack.config.dev.js` and `webpack.config.prod.js`), these are named explicitly in [scripts/starts.js](https://github.com/mikelax/forgingadventures/blob/issue-25/client/scripts/start.js#L31) and [scripts/build.js](https://github.com/mikelax/forgingadventures/blob/issue-25/client/scripts/build.js#L21) respectively.

Looking more into CRA README, it looks like only `NODE_ENV=production` and `NODE_ENV=development` are supported as these are explicitly set in the [build.js](https://github.com/mikelax/forgingadventures/blob/issue-25/client/scripts/build.js#L5) and [start.js](https://github.com/mikelax/forgingadventures/blob/issue-25/client/scripts/start.js#L5). `NODE_ENV=test` is available but it is only reserved for [running tests](https://github.com/mikelax/forgingadventures/blob/issue-25/client/scripts/test.js#L5) and not, for example, deploying to a test/qa server. For this, CRA recommends using [Custom Environment Variables](https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-custom-environment-variables) instead.

So if we wanted to deploy the UI to a test server, we cannot specify `NODE_ENV=test` in the compose file as that will be overwritten by `build.js`.

We could add another file to the `scripts` folder called `buildForTest.js` (for example) that then references `config/webpack.config.test.js`... something to think about when we come to deploy the UI to a test server.

For now, I've removed the environment setting from the client section of the docker-compose file - the api section still contains it for use in the API.

nazar:
This is also interesting: https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#advanced-configuration

mikelax:
thanks for all the research @nazar 
I guess what we can do is set up our build pipeline so when we run a "test" deploy that will deploy the "prod" client project and test API project. We'll just have to note that in the Build Readme. 

A "production" deploy will then deploy production for both the api and client.

mikelax (approved):

Issue #40, opened 2018-01-15
Infinite page scrolling for:


nazar:
1. Games list
2. Game messages

Ideally the same infinite page scroll component should be used for the above and any subsequent views that require it.

nazar:
Interesting article: https://medium.com/@gethylgeorge/infinite-scrolling-in-react-using-apollo-and-react-virtualized-graphql-cursor-pagination-bf80617a8a1a

mikelax:
I like that article, it seems to reference the cursor response structure similar to what FB Graph API does for indicating if there are more results to be returned for a given API call.

nazar:
The apollo client docs also have an example of cursor-based scrolling as well here: https://www.apollographql.com/docs/react/recipes/pagination.html#cursor-pages

I looked at several infinite scroll component but all their issues weren't inspiring. They either had issues with react 16, cross browser issues, performance issues, reverse paging issues or were pretty large (i.e. virtualised lists) for such simple functionality.

I did find an interesting one here: https://github.com/thebuilder/react-intersection-observer

It seems to use an almost widely used Browser API to detect intersection cheaply. The only browser that isn't currently natively supported is Safari, but I've updated the [webpack polyphills to load for unsupported browsers](https://github.com/thebuilder/react-intersection-observer#polyfill-for-intersection-observer). Currently testing that component (along with cursor paging 😅 )

nazar:
Closing in favour of #160

Issue #31, opened 2018-01-06
Create Game Lounge


mikelax:
Overview
===
As a user, after searching and finding a potential Game to Join, I will want to view and Post in the Game Lounge. This will be discussion about the meta of the Game, ie. potential players asking questions, posting character ideas, etc. This discussion will *not* be part of the Game itself. It will not be part of the IC or OOC, what we currently have implemented as the _game_messages_ table. 

Requirements
===
The Post "widget" should be the same widget that is used to make game messages pots. We may just want to disable a few of the toolbar options such as IC, Singing, Etc. 

The User must be logged in to create a Post in a Game Lounge. 

The messages will be displayed in chronological order

Questions
---
Should this be implemented as a sibling table to _game_messages_? ie. game_lounge? How would we make use of a commentable pattern that you mentioned a while back? Is that overkill for this use case?

nazar:
>Should this be implemented as a sibling table to game_messages? ie. game_lounge?

👍 

>How would we make use of a commentable pattern that you mentioned a while back? Is that overkill for this use case?

Yeah it might be overkill for this use-case as it would be simpler to implement a concrete table for this as opposed to a polymorphic table

>The Post "widget" should be the same widget that is used to make game messages pots. We may just want to disable a few of the toolbar options such as IC, Singing, Etc.

I was thinking of creating a ForumMessage component instead, which is also based on the draft editor. These two "message" components would be separate as the ForumMessage component would be easier to implement as draftjs already provides [richedit components](https://draftjs.org/docs/api-reference-rich-utils.html#content) (i.e. bold, italic, etc) leaving us just to implement a handful of specific ones - i.e. qoute block, which can reference the same quote block component used in the GameMessage component.

mikelax:
sounds good @nazar . I guess with the components being pretty small to begin with having a "duplicate" for the forum posting isn't a big deal as the actual code snippet is pretty small.

PR #41, opened 2018-01-15
refs #40 - implement infinite scrolling for /games list of games

nazar:
Hey @mikelax 

I've implemented simple offset based paging for now as I was struggling to figure out how to implement cursor based paging on the server. We can revisit this once more features are out of the way?


nazar:
refs #40

client/src/components/Games/queries.js
```diff
@@ -1,8 +1,8 @@
 import gql from 'graphql-tag';
 
 export const gamesQuery = gql`
-  query games($page: Int, $searchOptions: SearchOptions) {
-    games(page: $page, searchOptions: $searchOptions) {
+  query games($offset: Int, $searchOptions: SearchOptions) {
```
mikelax, line 4: @nazar  is this offset using a cursor or a true offset? 
shouldn't the cursor approach send the "id" of the last id from the previous request?

nazar: I'm not using cursors for this initial implementation as I was not able to find examples, specifically for how the server handles cursors. The example I've seen mention cursors on the client but not how there are generated on the server and sent to the client, specifically in relation to any existing search parameters.

I didn't want to spend too much time on this. We can revisit once this becomes a performance issue and we have implemented more features.

nazar:
Hey @mikelax - can I go ahead and merge this PR? I'd like to access the features in this branch for #31

mikelax:
@nazar sure. I will do it now.

PR #42, opened 2018-01-16
37 Implement update game messages

mikelax:
The graphql changes in the api project should be good. 
I am sure there could be some refactoring done in the react client project.


client/src/components/Games/View/GameMessages.js
```diff
@@ -45,7 +45,7 @@ class GameMessages extends Component {
     return <div className='game-messages'>
       {_.map(gameMessages, ({ id, message }) => (
         <div key={id} className='game-message'>
-          <GameMessage message={message} readOnly={true} />
+          <GameMessage message={message} id={id} gameId={this.props.gameId} readOnly={true} />
```
mikelax: @nazar  ?? should this be done another way with redux?

client/src/components/Games/components/GameMessage/index.js
```diff
@@ -1,28 +1,33 @@
...
-This module also contains GamesMessage sub-components - i.e. toolbar and inline and block style definitions
+/**
+ * GamesMessage is export wrapper component around draft js
+ * This module also contains GamesMessage sub-components - i.e. toolbar and inline and block style definitions
  */
-
-export default class GamesMessage extends Component {
+const GamesMessage = class GamesMessage extends Component {
 
   state = {
```
mikelax: @nazar  instead of creating a state object here, should this instead be done in a class constructor?
ie:

```js
  constructor(props) {
    super(props);
    this.state = {
      editorState: EditorState.createEmpty(decorator),
      editing: false,
      readOnly: this.props.readOnly
    };
  }
```

client/src/components/Games/components/GameMessage/index.js
```diff
@@ -1,28 +1,33 @@
...
-    editorState: EditorState.createEmpty(decorator)
+    editorState: EditorState.createEmpty(decorator),
+    editing: false,
+    readOnly: this.props.readOnly
   };
 
   componentWillMount() {
     const {message} = this.props;
 
+    // this.setState({readOnly});
```
mikelax: @nazar  I was setting the attribute in this function initially, but I don't think its right and I think this should be removed, right?

client/src/components/Games/components/GameMessage/index.js
```diff
@@ -46,26 +51,71 @@ export default class GamesMessage extends Component {
     }
   }
 
+  handleEdit = (editorState) => (e) => {
```
mikelax: @nazar  Do you think we should use this convention for function names that handle user interactions?

client/src/components/Games/components/GameMessage/index.js
nazar: no as `state = ...` is the short form of the constructor boilerplate code.

client/src/components/Games/components/GameMessage/index.js
nazar: Good question @mikelax. I'm still not sure about naming conventions as I do have concerns about react code readability in regards to how classes are structured. 

Ideally we want to see the propTypes, react lifecycle methods (i.e. `componentWillMount`) and the `render` methods at the top of the file to determine what the class does.

I'm not a big fan of the current convention where all methods in a class are public as it makes mentally parsing a class difficult. For example in this file the handlers preceed the `render` method.

I think it would easier if Javascript included support for private methods. In this absence I'm still toying with the idea that all support methods (such as handlers) should be defined as private functions outside the class.....

client/src/components/Games/components/GameMessage/index.js
```diff
@@ -75,9 +123,10 @@ export default class GamesMessage extends Component {
   getEditorMessage() {
     return convertToRaw(this.state.editorState.getCurrentContent());
   }
-
 };
 
+export default graphql(updateGameMessageMutation)(GamesMessage);
```
nazar: I'm not sure about this as it links the update mutator to the base GameMessage class. This should instead be handled by the parent component, like in the CreateMessage class.

I'll go ahead and refactor this.

nazar (commented):
Super thanks for this @mikelax - it is 95% there 👍 . I'll go ahead and refactor the react code and push shortly.

client/src/components/Games/components/GameMessage/index.js
mikelax: Right, no private functions yet .... 

@nazar if we move private functions outside the class does that though just end up making it more complicated to trace code later?  

What if we move all the “private” functions below render?  

So you have State definition (constructor), lifecycle functions, render, then finally private functions to handle user input, jsx helper functions, etc?

client/src/components/Games/components/GameMessage/index.js
mikelax: Ahh I wasn’t sure how to actually “link “ the user input submitnto making the graphql mutation call. 

Still have reading to do on Apollo and associated other libs we use.

client/src/components/Games/components/GameMessage/index.js
nazar: >if we move private functions outside the class does that though just end up making it more complicated to trace code later?

That depends on your editor :p

>What if we move all the “private” functions below render?

Yep - I think that is the best compromise for now

client/src/components/Games/components/GameMessage/index.js
nazar: >Ahh I wasn’t sure how to actually “link “ the user input submitnto making the graphql mutation call.

TBH same here! Looking into this now.

nazar: SO knows how: https://stackoverflow.com/a/43384358

nazar:
Pushed refactor

client/src/components/Games/queries.js
```diff
@@ -75,3 +75,13 @@ export const onGameMessageAdded = gql`
     }
   }
 `;
+
+export const onGameMessageUpdated = gql`
```
mikelax: @nazar  I see this subscription was added but it doesn't appear to have been referenced anywhere? Should the game messages list page also subscribe to it?

client/src/components/Games/components/GameMessage/index.js
```diff
@@ -76,7 +88,51 @@ export default class GamesMessage extends Component {
     return convertToRaw(this.state.editorState.getCurrentContent());
   }
 
-};
+  getFooter = () => {
+    const footerMessage = this.state.numberEdits > 0 ? 
+        <span>Number of Edits: {this.state.numberEdits}</span>
```
mikelax: @nazar  The only thing I wasn't sure about with this recent commit is how to get the number displayed here to refresh after the save has completed. I *think* it should be handled once you hook up the subscription, which would cause a render of the updated GameMessage component, and therefore the edit number will get updated? It doesn't feel right to update it in the submit function, for example.

nazar: Hey @mikelax 

>I think it should be handled once you hook up the subscription, which would cause a render of the updated GameMessage component, and therefore the edit number will get updated? It doesn't feel right to update it in the submit function, for example.

Yep, ideally this should be updated by the update subscription. I'll look into it 👍

api/Dockerfile
```diff
@@ -7,4 +7,4 @@ ADD . /app
 
 EXPOSE 3001
 
-CMD yarn && yarn migrate:latest && yarn start
```
nazar, line 10: hey @mikelax - I removed the `yarn migrate:latest` from the start script as it was crashing the api container when switching branches. I think you mentioned the same thing a while ago.

It might be less painful to run migrations manually for a while and only include the migration during deployments.

You'll have to run `docker-compose build` for the Dockerfile change to be registered.

As for running migrations manually, we'll need to shell into the api container using `bash -c "clear && docker exec -it forgingadventures_api_1 sh"` and run `yarn migrate:latest` in the container.

mikelax: thanks @nazar  I was actually running the migrations manually anyway so that change is fine with me.

client/src/components/Games/components/GameMessage/index.js
```diff
@@ -115,7 +115,11 @@ export default class GamesMessage extends Component {
   };
 
   handleCancel = (e) => {
-    this.setState({readOnly: true, editing: false});
+    this.setState({
+      readOnly: true,
+      editing: false,
+      editorState: EditorState.createWithContent(convertFromRaw(this.props.message), decorator)
```
nazar: ❤️

PR #45, opened 2018-01-18
game lounges

nazar:
Hey @mikelax,

This PR includes a basic game lounge forum that allows posting messages. It's currently implemented using tabs but it would be trivial to to move the game lounge to a seperate page - i.e. `/games/1/lounge`. I can do that on another story. This branch contains several changes so I'm keen to get it back into master to minimize merge conflicts.

I implemented the the game lounge messages, specifically to editing, slightly differently than how we implemented game message editing; I wrapped each gameLounge object [in a container](https://github.com/mikelax/forgingadventures/blob/31-game-lounges/client/src/components/Games/View/components/GameLoungeMessages.js#L102) that handles the editing state, controls and mutations - this really [simplified the lounge](https://github.com/mikelax/forgingadventures/blob/31-game-lounges/client/src/components/Games/components/GameLoungeMessage/index.js) editor component as it only needed the `message` prop, and not `numberEdits` and so on. It also [reduced the number of states](https://github.com/mikelax/forgingadventures/blob/31-game-lounges/client/src/components/Games/components/GameLoungeMessage/index.js#L9) the actual editor needed as well - i.e. no need for `readOnly`, for example as `readOnly` prop is now controlled by the parent component depending if the message is being edited.

 I'll go ahead and refactor the game message related components to this same structure on a seperate ticket.

Addresses #31


api/schemas/mutations.js
```diff
@@ -0,0 +1,16 @@
+const mutations = `
```
mikelax, line 1: @nazar  how come you moved the mutations, queries, and subscriptions into their own files rather than with the logical object they are for?

client/src/components/Games/View/components/assets/GameLoungeMessages.js.styl
```diff
@@ -0,0 +1,35 @@
+.game-lounge-messages
```
mikelax: @nazar  how come this filename extension is .js.styl?
other stylus files end in .styl

client/src/components/Games/View/index.js
```diff
@@ -16,21 +20,19 @@ class ViewGame extends Component {
 
   render() {
     return (
-      <React.Fragment>
```
mikelax: @nazar  I had used the Fragments so the Helmet code is separate from the actual DOM UI code.

mikelax:
@nazar  I see you added a new function to user service called `getOrCreateUserByAuth0Id`
I see its used in the schemascopegate. How does the `req.user.sub` get populated to be available in those functions? Is that something new with the new version of auth lock?

The `name` attribute is nullable in the users db, but the graphql schema marks the name attribute as required. Should we make those match? I had to manually add a value for the name column, this mismatch will prevent a new user from signing up and being able to post a lounge message. 

I know there is a note for collecting required fields after initial sign up, are you thinking we handle this logic when that note/issue is implemented?

mikelax:
@nazar question, is it correct that we have to use some type of [plugin like this](https://www.draft-js-plugins.com/plugin/static-toolbar) to get icons for the standard buttons like h1-6, bold, ul, ol, etc? 

This can be handled in separate ticket FYI.

https://github.com/draft-js-plugins/draft-js-plugins
*If* this support react 16, i see an open ticket for at least the emoji plugin then it might be viable. I would just like to get the standard buttons in, but if we can switch to a plugin that works and makes things easier than all the better. We would have to see how this would work with a custom toolbar for the game message icons we'll need.

mikelax:
also we get a `name` attribute from the [auth0 normalized profile](https://auth0.com/docs/user-profile/normalized/auth0). maybe we can default the name field in the db from the auth0 profile? It won't always be a "correct" naem, ie for twitch the `name` is just my twitch username, not full name.

api/schemas/mutations.js
nazar: Hey @mikelax - that's the [recommended implementation for graphQL schema stitching](https://www.apollographql.com/docs/graphql-tools/generate-schema.html#modularizing) for schemas for the same domain. The previous schema stitching was incorrect as we were not able to reference other types - i.e. not able to reference the User type in the GameLoungeMessage type, for example. That implementation was for stitching remote schemas.

nazar:
Hey @mikelax 

>I see you added a new function to user service called getOrCreateUserByAuth0Id
I see its used in the schemascopegate. How does the req.user.sub get populated to be available in those functions? Is that something new with the new version of auth lock?

I forgot to mention that yesterday! I didn't want to setup a tunnel on my dev (i.e. extra dev step) on my dev for the auth0 login callback so I instead I experimented with [querying the user details from auth0](https://github.com/mikelax/forgingadventures/blob/31-game-lounges/api/services/user.js#L33) whenever we need to find a user from their token, specifically the user `name` and `picture`. This is a one time auth0 API query that fires when the user's auth0 ID doesn't exist in our shadow `users` table.

>The name attribute is nullable in the users db, but the graphql schema marks the name attribute as required.

Yeah, I think the `name` attribute should be `NOT NULL` in the DB as it is required for GameLounge posts.

>question, is it correct that we have to use some type of plugin like this to get icons for the standard buttons like h1-6, bold, ul, ol, etc?

Definitely!! DraftJS does have some amazing plugins (like the one you mentioned). There is also a [collection of editor implementations](https://github.com/nikgraf/awesome-draft-js#standalone-editors-built-on-draftjs) based on DraftJS. We definitely should be able to use any of these for the GameLounge Post editor as that component's internal implementation can easily be switched out without affecting the parent components.

>also we get a name attribute from the auth0 normalized profile. maybe we can default the name field in the db from the auth0 profile? It won't always be a "correct" naem, ie for twitch the name is just my twitch username, not full name.

I believe that's what I'm doing [here](https://github.com/mikelax/forgingadventures/blob/31-game-lounges/api/services/user.js#L33) - plus the user should be able to edit any of these in the Account Profile page as described in the *Convert Profile Page to User Account Settings Page* note.

mikelax (approved):
🚀

Issue #48, opened 2018-01-23
Add game label to Games table, Game List, Game Details


mikelax:
Overview
===
Now that the table structure around the database **games** objects are solidifying we can add the concept of game labels. The table already exits, we just need to add the foreign key to the games table, update apollo types, and then update the Games UI elements. 

Requirements
===
Add a foreign key from game_labels and games tables. 
Update apollo types for games to include labelId and game_labels object. 
* On /games page, add a Game Labels to search form. For v1 it can just be a regular dropdown, no need for auto-complete. 
** When displaying game results, display the game label for each game
* On /games/:id page, display the game label for the game
* On /games/create add a new field for Game Label. For v1 it can be a basic dropdown

mikelax:
Look at using the [Ribbon element](https://react.semantic-ui.com/elements/label) for adding the game label to the cards.

Issue #30, opened 2018-01-06
Add support for Applying & Joining a Game


mikelax:
Overview
===
As a user, after reading the details of a Game, and possibly posting in the Game lounge, a user will want to Join a Game. As part of making a Post in the Lounge, a User will be able to apply to join the Game. The user will optionally be able to enter a message, but at the very least the act of applying will create a message such as:  _Michael has applied to join the Game!_

This ticket will mainly cover the DB schema, methods and use cases for joining and the GM accepting players. A (basic) prototype UI can be added to the Game Details / Lounge Page. 

Requirements
===
A user may apply to a Game, this will put the User in the initial state of:  "**pending**".
The GM of the Game may then update the status of the User to:  "**accepted**", or "**rejected**".
Though unlikely, a User may apply to the same game more than once.

A User must be logged in to create a Post in the Game Lounge and Join a Game. If the User is not logged in and clicks the Join button they should be redirected to the Login page.

Technical Details
---
For this ticket we will only handle the use case of a User applying to a game. A future enhancement to this ticket will allow for the User to optionally select a new or existing character to include when joining the game. 

The db needs to support the use case of a user joining a game with a nullable (blank) character. The character can be filled in at a later point in time. 

Database Tables
---
Here is a list of possible changes to support this feature. 

**game_players**
id, userId, createdAt, status (pending, accepted, rejected, quit, kicked), updatedAt, deletedAt
Once we add the concept of characters, a nullable characterId will be added. 

UI Implementation
---
In this ticket we will implement a basic UI to all for the following use cases:
* Joining a Game
* Viewing the status of Players in Game
* Allow GM of a Game to accept a Player

We will add a new tab on the View Game page between the existing two existing Tabs. The new tab name will be called:  "Players"
The new tab will show a table of the players in the game that have a status of pending and accepted. Rejected and kicked status, etc players will not be displayed. 
Above the table will display the number of min, max, and current accepted players.
The table will display the following information: 
User Profile Image, Name (Timezone if set up already)
Status
In a future ticket we will add character information such as:  Profile Image, Name, Level, Race, Class

Join Game
---
A new button will be added to the View Game "header" where the game title, scenario is displayed above the tab bar. The button should be right aligned. The label should be:  **Join Game**
It should change to a disabled state of "Joined"  if the user is currently accepted or the game-master. 
It should be a disabled state with a label of "Pending" if the user is pending. All other states (ie. rejected, etc) should display the default **Join Game**.

[...]

PR #49, opened 2018-01-25
30 Join Game

mikelax:
Still very much a WIP. 

**NOTE** - the migration in this PR will delete data from all the games, messages, and lounge messages tables to allow for adding the foreign key to game_labels. This was to prevent having to drop this data later on once ready to work on #48.

### DIFF
--- a/api/schemas/gamePlayer.js
+++ b/api/schemas/gamePlayer.js
@@ -1,0 +1,100 @@
+import { withFilter } from 'graphql-subscriptions';
+
+import GamePlayer from 'models/gamePlayer';
+import schemaScopeGate from 'services/schemaScopeGate';
+import { getOrCreateUserByAuth0Id } from 'services/user';
+import pubsub from 'services/pubsub';
+
+export const GAME_PLAYER_ADDED = 'game_player_added';
+export const GAME_PLAYER_UPDATED = 'game_player_updated';
+
+export const gamePlayerTypeDefs = `
+
+  type GamePlayer {
+    id: ID!,
+    user: User!,
+    gameId: ID!,
+    status: String!,
+    updated_at: GraphQLDateTime,
+    created_at: GraphQLDateTime
+  }
+  
+  input CreateGamePlayerInput {
+    gameId: ID!,
+    status: String!
+  }
+  
+  input UpdateGamePlayerInput {
+    status: String!
+  }
+`;
+
+export const gamePlayerResolvers = {
+  GamePlayer: {
+    user: gameLoungeMessage => User.query().findById(gameLoungeMessage.userId)
+  },
+  Query: {
+    gamePlayer: (obj, { id }) => GamePlayer.query().findById(id),
+    gamePlayers: (obj, { gameId }) => GamePlayer.query().where({ gameId }).orderBy('created_at')
+  },
+  Mutation: {
+    createGamePlayer: (obj, { input }, context) =>
+      schemaScopeGate(['create:games'], context, () => {
+        return getOrCreateUserByAuth0Id(context.req.user.sub)
+          .then((user) => {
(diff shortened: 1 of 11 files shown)


client/src/components/Games/queries.js
```diff
@@ -136,6 +148,17 @@ export const onGameMessageUpdated = gql`
   }
 `;
 
+// game players
+
+export const gamePlayersQuery = gql`
+  query gamePlayers($gameId: ID!) {
```
mikelax: @nazar  I tried to add in a parameter for status, but I couldn't get the syntax to work, was getting an HTTP 400 error.

client/src/components/Games/queries.js
```diff
@@ -42,6 +42,18 @@ export const gameLoungeMessageFields = `
   ${userMetadataFields}
 `;
 
+// game polayers
+
+export const gamePlayerFields = `
+  fragment gamePlayerFields on GamePlayer {
+    id
+    user {
```
mikelax, line 41: @nazar  these can be updated once we merge in the ticket you are working on

api/schemas/gamePlayer.js
```diff
@@ -0,0 +1,104 @@
...
+              .tap((gamePlayerMessage) => {
+                pubsub.publish(GAME_PLAYER_UPDATED, {
+                  gamePlayerUpdated: gamePlayerMessage
+                });
+              });
+          });
+      })
+  },
+  Subscription: {
+    gamePlayerAdded: {
```
mikelax, line 89: @nazar  I'll need some help with hooking up the subscription to the players component.

client/src/components/Games/View/components/GamePlayers.js
```diff
@@ -0,0 +1,32 @@
...
+          ))}
+        </div>
+      </React.Fragment>
+    );
+  }
+}
+
+export default compose(
+  graphql(gamePlayersQuery, {
+    options: ({ gameId }) => ({ variables: { gameId, status: ['accepted'] } })
```
mikelax: @nazar  once we get the other syntax error working in the query, then the status here should be updated to also include `'pending'`

client/src/components/Games/queries.js
nazar: @mikelax - did you add the `status` parameter to the [query definition](https://github.com/mikelax/forgingadventures/blob/30-joingame/api/schemas/queries.js#L21) in the API schema? 

I'm thinking maybe

```
gamePlayers(gameId: ID!, status: [String]): [GamePlayer!]
```

I'll have a play with the query

nazar: yep - that worked

mikelax: thanks @nazar  I pulled and I see the query now returns, no 400 error. 
One issue though is that it does not appear to respect the values I submit here. 

ie. I think the destructured "default" values I am using [here](https://github.com/mikelax/forgingadventures/pull/49/files#diff-837acce59d7ca64e0718c0c0f8229800R39) are always being used for the query, even if a value is passed in to the `status` parameter.

nazar: @mikelax the `status` parameter will also have to be defined in the [client query](https://github.com/mikelax/forgingadventures/blob/30-joingame/client/src/components/Games/queries.js#L154-L155).

I tried a manual query from graphiql and it did accept the status when I set it in the query,

mikelax: @nazar  that worked. thanks!  🚀

nazar: 🚀  🎆  🚒  👨‍🚒

mikelax (commented):
does not include joining game, but displays game players.

mikelax:
Addresses #30

Issue #43, opened 2018-01-16
Make UI handle expired JWT Tokens


mikelax:
Overview
===
There is currently a bug on the UI as to how it handles expired JWT tokens. If the UI thinks the user is still logged in (ie. cookies exist) but the token in the cookie is expired then currently the game messages errors out. This is due to it making graphql calls with the expired token instead of redirecting to the login page. 

Requirements
===
I have run into this use case leaving my browser open. In this case I *think* the system that renews the token by hitting the /silent page should handle this, but there might be a bug in that logic. 

Also, we probably should also have the UI itself set up to handle receiving *unauthorized* HTTP responses and redirecting back to the login page. 

<img width="1214" alt="screen shot 2018-01-16 at 4 07 15 pm" src="https://user-images.githubusercontent.com/2080701/35012448-c8adca26-fad7-11e7-8ae3-7b358df129e6.png">

mikelax:
hey @nazar  I think this issue is still happening. I just checked out your new 53 branch to test it, the dockers containers were already running so they recompiled because of the changes. My web browser was already on the /games page, after reloading I received the same error I had been receiving when there is a jwt token in the browser storage but its expired. 

Here is stack trace: 
```
api_1     | UnauthorizedError: jwt expired
api_1     |     at /app/node_modules/express-jwt/lib/index.js:102:22
api_1     |     at /app/node_modules/jsonwebtoken/verify.js:27:18
api_1     |     at _combinedTickCallback (internal/process/next_tick.js:131:7)
api_1     |     at process._tickDomainCallback (internal/process/next_tick.js:218:9)
```

Here is web browser stack trace
<img width="1159" alt="screen shot 2018-01-27 at 8 08 01 am" src="https://user-images.githubusercontent.com/2080701/35472240-70bf6a62-0339-11e8-97f0-15a1d2d4bc0f.png">

nazar:
I'll re-open 👍

nazar:
I'm seeing this error when the UI attempts to re-authorise a token - digging deeper

<img width="1412" alt="screen shot 2018-01-27 at 13 56 36" src="https://user-images.githubusercontent.com/6967/35472608-100eead6-036a-11e8-8529-6fc46d1ecec5.png">

mikelax:
hmm, maybe that is related to helmet security installed on the express server?

nazar:
Yeah, I thought it was helmet as well but I was seeing the same message with the helmet middleware commented out.

[...]

Issue #47, opened 2018-01-21
**Collect Required Fields** after initial User Sign up


mikelax:
Overview
===
As a User, after I complete the initial Sign up process, I should be redirected to a page to fill in a few additional required fields needed for the FA website. 

Users will continue to be redirected to this special data collection page after each log in action until the required fields are submitted, which will trigger an a flag to indicate the process is complete and the page can be skipped in future log in actions. 

Requirements
===
The URL for this page will be:  /login/almost-finished
If the User visits this page and they are not logged in, they should be redirected to /login
If the User visits the page and they are logged in, but `signupCompleted` is true, they should be redirected to /games (eventually this may change once we have the proper account page created)
If the User visits the page and they are logged in, but `signupCompleted` is false is undefined, they they should stay on the page

Data Storage
---
Here are the fields that will be displayed on this additional sign up page:

* username (string) - required, globally unique
* name (string) - required
* profileImage (JSON) - optional
* timezone (String) - optional - will store the [IANA timezone name](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) ie. America/New_York

The storage for these fields will be in the users postgres table. 

Logic
---
After the data has been saved to the users table, the function `auth0.patchAuth0Metadata` should be called to push the user and app metadata fields to Auth0. 

**NOTE** - after thinking about this more I do not think that we need to store *every* profile attribute value in the Auth0 user_metadata field. There really isn't a point to this, we should only store values that might be needed by Auth0 functions, ie. sending emails, etc. I can think that only the username, name, and profileImage (url) will be required). This also makes it easier if we decide to add additional fields to our user profile in the future, we don't have to worry about pushing these fields to Auth0. 

The user metadata should contain:  
* username (string)
* name (string)
* profileImage (string) - will just be the "basic" https image url, not a json object with the publicId

The app metadata should contain:  
* faUserId
* signupCompleted

Sign Up / Log in
---
After the initial sign up action, the user will be redirected to this page. 
After a log in action, if the attribute `signupCompleted` is not set to true, then the user will be redirected to this page. 

Once all required fields are submitted by the user, the field `signupCompleted` will be set to true in the users table. This field will be added to the `app_metadata` field.

Auth0 Data Sync
---
[...]

PR #52, opened 2018-01-26
#47 complete signup

nazar:
References #47


nazar:
also closes #43

api/schemas/user.js
```diff
@@ -1,25 +1,54 @@
...
-    user: (obj, { id }) => User.query().findById(id)
+    me: (obj, obj2, context) => getOrCreateUserByAuth0Id(context.req.user.sub)
+  },
+  Mutation: {
+    updateMe: (obj, { input }, context) => updateUserAndAuth0(input, context.req.user.sub),
+    validUsername: (obj, { username }, context) => {
+      return User
+        .query()
+        .count('username')
+        .where({ username })
```
mikelax, line 48: @nazar  do you think that we need to "normalize" the username
ie. lowercase, remove spaces and special chars to turn into url slug?

mikelax (approved):
looks really good! 
there are a couple of items we can discuss as separate tickets.

api/schemas/user.js
nazar: hey @mikelax - I was thinking about that as well and was looking into validation libraries that would ease the UI/API rule checking on usernames (and names) - alpha only, min and max length etc....

In the past I've used something like [checkit](https://github.com/tgriesser/checkit). Unfortunately that library has been abandoned (along with bookshelfjs - same author). Another alternative I looked at was [validate.js](https://validatejs.org/). Another alternative is [Joi](https://github.com/hapijs/joi), but that is a rather chunky library and I'm not too fond of it.

In the end I left validation out of the PR but it is something we should revisit 👍

auth0/rules/update-profile-with-custom-profile.js
```diff
@@ -0,0 +1,17 @@
...
+      user.name = user.user_metadata.name;
+    }
+
+    if (user.user_metadata.username) {
+      user.nickname = user.user_metadata.username;
+    }
+  }
+  
+  callback(null, user, context);
+}
```
nazar, line 17: hey @mikelax - I didn't know the auth0 user 'name' or 'picture' could be updated. I would have done that from the API - I should have checked the docs.

PR #54, opened 2018-01-27
closes #53 - Prevent submitting empty game lounge/message

### LINKED ISSUE #53: Prevent submitting empty game lounge/message
Overview
===
As a user, I should not be able to submit an empty game lounge message or game message 
Currently there is no validation on the submit button, so pressing it will create a new message. 

The button should have a check to be disabled if the textarea is in the default state (ie. no content).

nazar:


mikelax (approved):
🚀

PR #56, opened 2018-01-28
closes #43 - use newer weAuth.checkSession method that doesn't requi…

### LINKED ISSUE #43: Make UI handle expired JWT Tokens

nazar:
…re silent callback.


nazar:
Hey @mikelax - token renwals only work with email signups on dev. We'll have to test social signups at some stage on a non dev auth0 client. I removed the api silent.html page (and ejs) as new webAuth renwal method doesn't require the silent callback api page.

nazar:
Speaking of security - Black Mirror S03E06, Hated in the Nation, is well worth rewatching 🙀

mikelax:
hey @nazar  thanks for the work on this. 
I think that the twitch oauth client should work as that is using real api keys, currently only google is using dev oauth keys.

mikelax:
@nazar  hmm something still doesn't seem correct with this user flow. 

I see the call to auth0 /authorize. 
ie:
```
https://forgingadventures-staging.auth0.com/authorize?client_id=i7Srpetd9dL9Dx1ckMIhBLFnSxksaVcz&response_type=token%20id_token&redirect_uri=http%3A%2F%2Flocalhost%3A3000%2Fcallback&audience=https%3A%2F%2Fapi-test.forgingadventures.com&scope=openid%20profile%20email&connection=Username-Password-Authentication&timeout=5000&state=1tFzkJh4ATX2._xSE.~TZFIkGJSc-FD5&nonce=51i5QfBldZLkwFIsJeZ46tr4wu8qmTi7&response_mode=web_message&prompt=none&auth0Client=eyJuYW1lIjoibG9jay5qcyIsInZlcnNpb24iOiIxMS4wLjEiLCJsaWJfdmVyc2lvbiI6IjkuMC4xIn0%3D
```
and this appears to return a valid response and token. The problem is that the call to /graphql is still failing. I think this is because the browser still has the "old" token in the cookie storage and is not using the new one that was received from the /authorize call.

nazar:
Hey @mikelax - I just retested this with an email login user. The `webAuth.checkSession` call is made which returns the token result which is also set in localStorage. I can also see subsequent API calls being made using the new token.

I tested this by manually setting the scheduleRenewal delay to 3s. Maybe this doesn't work when the token actually expires? I'll test again with an expired token; it could be that we are not handling the expired token state correctly - i.e. need to redirect to the login screen as I noticed that in [this code section](https://github.com/mikelax/forgingadventures/blob/master/client/src/services/login.js#L45-L53) the `lock.on(*)` don't fire unless the UI is on the callback page with the correct callback query strings.

mikelax:
Ok cool I will give it a try this morning. 
Theoretically if I switch to this branch,  then on my first page request my token should be renewed?
Will this be true if the first page is the homepage, or one that doesn’t make a graphql call?

nazar:
>Theoretically if I switch to this branch, then on my first page request my token should be renewed?

That's what I'm not sure of. Currently [this block](https://github.com/mikelax/forgingadventures/blob/master/client/src/services/login.js#L40-L56) checks if you have a valid token. If not it goes into the `lock` related else block, which won't fire any of the `.on(*)` events unless it is on the callback page with the QS params from oAuth. 

What I've verified is that the timer based `renewToken()` works _IF_ you visit the app with a valid token _AND_ it expires whilst you have the app window open.

I have a feeling we need a third logic branch - i.e. have token but it is expired to then attempt to renew it.

mikelax:
@nazar  ok here are the test cases that I just ran through after switching to this branch:

1. I opened a browser to /games. I already had an expired token in the browser storage. 
I received graphql errors on the page. A call was made to auth0 /authorize and this response did return a valid access_token valid with a future exp date. 

2. I next opened a browser to the homepage. I still had the expired token in the browser storage. This did not return any graphql errors since none are made on the page. The same call was made to /authorize and that still returned a valid token. The browser application local storage though still contained the old expired access_token.

mikelax (approved):
this is looking really close. 
I think we just have the one last use case you mentioned. I will merge this in though as this is 95% complete.

PR #57, opened 2018-01-28
30 Add styling to game players tab.

mikelax:
Add styling to game players tab.

To Do List
---
- [x] Change button label / disable button on /game/:id page based on user's status in game
- [x] Pass `game` object in props to JoinGame component
- [ ] Call multiple mutate functions in _submit function in JoinGame. This is needed to insert the message after the game player mutation
- [x] Error handling in mutation?? ie. if player is already pending, they should not be able to join again.

### DIFF
--- a/client/src/components/Games/View/components/assets/GamePlayers.styl
+++ b/client/src/components/Games/View/components/assets/GamePlayers.styl
@@ -2,14 +2,16 @@
   display grid
   grid-template-columns 2fr 5fr
   grid-column-gap 30px
+  margin-top 15px
   margin-bottom 15px
+  align-items center
 
   .user-image
     img
       width 70px
-      border-top-left-radius 3px
-      margin-right: 5px
+      border-radius 50%
+      margin-right 5px
     .glyphicon-user
       width 70px
       height 70px
-      margin-right: 5px
+      margin-right 5px
--- a/client/src/components/Games/View/components/assets/GamePlayers.styl
+++ b/client/src/components/Games/View/components/assets/GamePlayers.styl
@@ -1,0 +1,15 @@
+.game-player
+  display grid
+  grid-template-columns 2fr 5fr
+  grid-column-gap 30px
+  margin-bottom 15px
+
+  .user-image
+    img
+      width 70px
+      border-top-left-radius 3px
+      margin-right: 5px
+    .glyphicon-user
+      width 70px
+      height 70px
+      margin-right: 5px
--- a/api/schemas/gamePlayer.js
+++ b/api/schemas/gamePlayer.js
@@ -41,7 +41,15 @@ export const gamePlayerResolvers = {
     gamePlayers: (obj, { gameId, status = ['pending', 'accepted'] }) => GamePlayer.query()
       .where({ gameId })
       .whereIn('status', status)
-      .orderBy('created_at')
+      .orderBy('created_at'),
+    myGamePlayer: (obj, { gameId }, context) => {
+      return getOrCreateUserByAuth0Id(context.req.user.sub)
+        .then((user) => {
+          return GamePlayer
+            .query()
+            .where({ gameId, userId: user.id });
+        });
+    }
   },
   Mutation: {
     createGamePlayer: (obj, { input }, context) =>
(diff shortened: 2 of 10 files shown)


client/src/components/Games/View/components/assets/GamePlayers.styl
```diff
@@ -0,0 +1,15 @@
+.game-player
+  display grid
```
nazar, line 2: ❤️ 👍

nazar (approved):
💯

client/src/components/Games/View/components/assets/GamePlayers.styl
mikelax: haha, I think you might like that. 
I was wondering @nazar  if you could take 10 minutes to try and improve the layout between the profile image, name, and to be sure the "status" column stays lined up. 

then the only thing we would need to refactor in the future is once we add support for the Character, which would be either a "Pending", N/A (for game master), or the actual character (image and name) selected.

nazar: I'm currently giving semantic-ui a twirl. This example looks relevant: https://react.semantic-ui.com/elements/list#list-example-image

mikelax:
@nazar  I updated this PR with some basic support for the /join page. 
There are a few concepts that I am not sure how to implement right now, they are the checkboxes in this PR. The main question is how to get props to the JoinGame component (without having the page make a graphql call). Also how to get the _submit function to call multiple mutations. Also need to know if refretchQueries is required. I *think* it will be as the user's local cache will need to be updated to get the new game player list, unless that will refresh when the user is redirected there.

nazar:
>There are a few concepts that I am not sure how to implement right now

Thanks for the work on this @mikelax - I can have a look at the questions raised and take it from here.

>Also how to get the _submit function to call multiple mutations

In terms of calling multiple mutations, these need to be defined in the `grapql` wrapper, like [here](https://github.com/mikelax/forgingadventures/blob/master/client/src/components/Games/View/components/GameLoungeMessages.js#L225-L227). Multiple mutations can be define like this, each with their own name which can then be [called from the props](https://github.com/mikelax/forgingadventures/blob/master/client/src/components/Games/View/components/GameLoungeMessages.js#L172).

mikelax:
thanks @nazar. I think I need to read up a bit more on thunk, as I saw some of the functions used elsewhere were related to that project. So I am still trying to work out how a page, ie. in this case "JoinGame" would get the info it needs, ie. the `game` object.

nazar:
>So I am still trying to work out how a page, ie. in this case "JoinGame" would get the info it needs, ie. the game object.

I haven't had a chance to have a good 👁  at the PR yet but one way to get the `game` object is to use a [graphQL query](https://github.com/mikelax/forgingadventures/blob/master/client/src/components/Games/View/index.js#L45-L47), like we do with the [View component](https://github.com/mikelax/forgingadventures/blob/master/client/src/components/Games/View/index.js#L55).

nazar:
>without having the page make a graphql call

My understanding of graphQL is it should used in any component that requires data. The graphQL client is clever enough to figure out if the requested object is in the cache or if it needs to make a call to the API server. We can also use the `recompose` library to add multiple queries, mutations, etc to a component like [here](https://github.com/mikelax/forgingadventures/blob/master/client/src/components/Games/View/index.js#L44-L50).

For the JoinGame component, [withRouter](https://github.com/mikelax/forgingadventures/blob/master/client/src/components/Games/List/index.js#L97) can also be added to the `compose` function to supply routing props (i.e. `match` to extract the `:id`). Having said that, the JoinGame component should already have routing props as it has been defined as a route Component [here](https://github.com/mikelax/forgingadventures/blob/30-joingame2/client/src/components/Games/index.js#L21).

nazar:
Hey @mikelax - please feel free to assign this ticket over to me once you've had enough :D

mikelax:
haha, @nazar  if you are going to take a peak at it tonight feel free to grab it now. 
I was going to try to wire up connect to add in the additional graphql query, but you can work on it. and then i'll review the changeset.

nazar:
hehe - I'll most likely have a look tomorrow morning. Amanda came back 30 mins ago and going to catch up with her.

mikelax:
@nazar  I am going to merge this PR in as well. 
We have the main use case covered. We will address changes the textarea to the main editor in a separate PR that refactors it to a RichEditor component.

mikelax:
Addresses #30

PR #65, opened 2018-02-07
closes #58 - Add ability to attach an image to a game

### LINKED ISSUE #58: Add ability to attach an image to a game
Update the create game view to accept an image. Refactor the users API endpoint to accept both profile and game images.

nazar:


api/migrations/20180207114450_58_user_uploads_extra_types.js
```diff
@@ -0,0 +1,17 @@
+exports.up = function(knex) {
+  return knex.schema.raw(`
+    ALTER TABLE "user_uploads"
+    DROP CONSTRAINT "user_uploads_type_check",
+    ADD CONSTRAINT "user_uploads_type_check" 
+    CHECK (type IN ('userProfile', 'characterProfile', 'gameImage', 'messageImage'))
```
mikelax, line 6: @nazar  should we change messageImage to messageAttachment so its general to support video types?

mikelax:
@nazar this looks pretty good, one question though. 
Now that it is a general image upload endpoint, I didn't know where in the logic an image is saved to the "users" table for a user profile image vs. the games table for game cover image, etc.

api/migrations/20180207114450_58_user_uploads_extra_types.js
nazar: The only issue with that is it might be tricky picking [transformations](https://github.com/mikelax/forgingadventures/blob/58-game-img/api/services/cloudinary.js#L9-L13) depending on the type if both photo and video share the same type names.

Having said that, that might not be an issue if we decide to go ahead with the Cloudinary React SDK for the UI.

nazar:
>Now that it is a general image upload endpoint, I didn't know where in the logic an image is saved to the "users" table for a user profile image vs. the games table for game cover image, etc.

Good question Mike. The new image API endpoint is purely for image uploads. The updates to related tabes (i.e. `users.userProfile`, `game.gameImage` is done in the UI [here](https://github.com/mikelax/forgingadventures/blob/58-game-img/client/src/components/Login/AlmostFinished/index.js#L197-L211) and [here](https://github.com/mikelax/forgingadventures/blob/58-game-img/client/src/components/Games/Create/index.js#L158-L168).

We went for this two step approach as Apollo and graphQL 1.x doesn't have well agreed on support for the File type to use in graphQL schemas.

mikelax (approved):
🚀

PR #71, opened 2018-02-10
68 Only allow users to edit their own messages

### LINKED ISSUE #68: Only original user should be able to edit message
Overview
===
Currently any logged in user is able to edit a lounge or game message. this should be refactored so only the user that created the message should be able to edit.

mikelax:
Fixes #68 
@nazar  I was able to use the `sub` attribute from the api access token to get the user id from the db to compare ownership to message or lounge message. I think this is better than having the UI also submit the id, as that is not as secure.


mikelax:
@nazar  one question in going over the gameLounge mutation. [What is this line](https://github.com/mikelax/forgingadventures/blob/f8bd5d8fbfefc93db163eb624479f29a61e94034/api/schemas/gameLounge.js#L51) used for, as far as I can tell it's not currently being used, but was added "in case" the id was needed for future logic?

nazar:
Hey @mikelax 

>one question in going over the gameLounge mutation. What is this line used for, as far as I can tell it's not currently being used, but was added "in case" the id was needed for future logic?

Yeah, that line doesn't make sense as we would never need to change the message userId on an update. 👍 to remove that line

nazar (approved):
💯  👍  :shipit:  🚀

mikelax:
thanks. will do!

Issue #60, opened 2018-01-31
Use RichEditor component when joining Game


mikelax:
Overview
===
After the new RichEditor has been created in #59 it should replace the existing textarea on the Join Game page. The editor will have some default text loaded, but the user will be free to add additional text to the message. 

Requirements
===
Replace the textarea with the new RichEditor.  There is a  `todo` in the /Join/index.js to address this issue. 
The editor should have the default text of:  
_Name has applied to join the Game!_

Refactor the _submit function to also call the `createGameLoungeMessageMutation` mutation to create the new lounge message after the `createGamePlayer` mutation.

Issue #59, opened 2018-01-31
Refactor game lounge message into reusable RichEditor component


mikelax:
Overview
===
We want to refactor the existing editor that is currently used on the game lounge messages tab into a re-usable component. 

Requirements
===
This component should be built in such a way that it can be used in any generic use case. This even includes using it outside the scope of a game. An example of this would be if we want to allow for a "rich" user profile, or for the purpose of the user requesting to join a game. 

The editor should be able be to be instantiated with a blank textarea, specific ghosttext, or an actual  default message. 

As part of this change the existing editor on the game lounge messages tab should be changed to use this new component. A separate ticket will cover the work to use the component on the join game page.

PR #72, opened 2018-02-10
Rich Editor Refactor

nazar:
References #59 and #60


mikelax:
@nazar  super excited to try this PR! code looks nice. i assume the meta column added to lounge messages is for the "default" text we may want to add? 

If things work in this branch is it ok to merge?

mikelax (approved):
🚀  nice job, looks good!

nazar:
Hey @mikelax - morning!

>@nazar super excited to try this PR! code looks nice. i assume the meta column added to lounge messages is for the "default" text we may want to add?

I was planning to explain the `meta` column but I was short on time over the weekend. Although I did play around and implement a method in which an inline styled text can be injected into the editor on the page load the result was rather flakey - i.e. I had to wait for the form to settle re-rendering from all the graphQL props before I could get the editor's ref and inject the text.

So I thought more about what you'd mentioned in regards to having a status text outside of the editor and implemented the `meta` column for that. For the moment this only allows the `join` status but this will later be expanded by `resign`, `kick`, `reject` etc. The message lounge component [checks the meta values and renders these status blocks](https://github.com/mikelax/forgingadventures/blob/master/client/src/components/Games/View/components/GameLoungeMessages.js#L151-L163) - right now it looks basic but we can always style these better (i.e. with colours and icons).

Issue #64, opened 2018-02-06
Replace bootstrap dropdown with more general select component


mikelax:
Overview
===
Now that we have replaced the UI widget layer from bootstrap to semantic in #62 , we need to remove the final lib that is used on the almost finished page. This lib `react-bootstrap-timezone-picker` should be replaced with [react-select](https://github.com/JedWatson/react-select).

Requirements
===
Remove the `react-bootstrap-timezone-picker` and sibling `react-bootstrap` libs from the client.
Obtain a JSON file structure that contains all the standard timezones, that will store the tz data in the same format as the previous lib. 

[Here is an example](https://github.com/JedWatson/react-select/blob/master/examples/src/components/CustomRender.js#L32-L50) where you can use the select component as a multi-select with custom rendering for the selected items, to be able to instead display closer to a material design "chip".

mikelax:
hey @nazar  sunday is looking like a total washout of a day. rain all day so I will have some time to work on this ticket, and hopefully more. Any additional thoughts besides trying to implement with the react-select ? Only real question will be finding the timezone json file, but otherwise should be pretty straight forward. 

We'll also need to make sure that we can lookup from a value, ie: America/New_York and get back to the correct value in the dropdown, and also to be able to get the GMT hour offset. so we have the option to display the +- GMT offset on the game players tab.

PR #73, opened 2018-02-11
64 refactor timezone select into component

mikelax:
Remove bootstrap libs, replace with react-select. 
refactored into component. 

*only* potential UI issues is the light border around the input select box, the div with the  "Select-input" class.


nazar (approved):
💯  👍  🚀  :shipit:

mikelax:
Closes #64

Issue #63, opened 2018-02-06
Cleanup Client login logic


nazar:
Related to #43.

Tasks

1. Refactor the top level client `login`, `callback` and `almost-finished` routes into a single `login` view with the `callback` and `almost-finished` as sub-routes.

2. Refactor [this](https://github.com/mikelax/forgingadventures/blob/master/client/src/services/login.js#L44-L54) block of code to only be called in the new `login/callback` component as currently the Promise in the block of code never fulfils as the `lock.on` events only trigger durring the auth0 callback. Instead attempt to use `webAuth.checkSession` to auto login the user if their auth0 session is still active.

PR #76, opened 2018-02-13
63 - Clean up client login logic

nazar:
Hey @mikelax 

For this PR please change your client `.env.local` `REACT_APP_AUTH0_REDIRECT_URI` to the following:

```
REACT_APP_AUTH0_REDIRECT_URI=http://localhost:3000/login/callback
```

I've also added the `http://localhost:3000/login/callback` route to the `i7S...` clientId setting under auth0 -> Clients -> **Allowed Callback URLs**. We can remove both the `http://localhost:3000/callback` and `http://localhost:3001/silent` urls from the **Allowed Callback URLs** after this PR.

I also added `http://localhost:3000` to **Allowed Logout URLs** as the UI logout functionality now also logs out the user from their auth0 session as well.

Refs #63


client/src/components/Login/index.js
```diff
@@ -1,61 +1,23 @@
...
-            </Header>
-
-            <div id="auth0Lock"/>
-          </Container>
-        </div>
-      </React.Fragment>
+      <div className="login">
+        <Switch>
+          <Route path={`${match.url}/callback`} component={Callback}/>
+          <Route path={`${match.url}/almost-finished`} component={AlmostFinished}/>
```
mikelax: do we need to add the `AuthGuard` component? 
They were there prior, and now i am able to get to this route not logged in.

mikelax (approved):

mikelax:
Closes #63

Issue #80, opened 2018-02-19
Update Menu and replace Logout with dropdown menu


mikelax:
Overview
===
After a user login replace the logout button with the user image and username with drop down menu to include entries:

* Profile
* Divider
* Logout

We will then be able to add additional options such as Account, Games, etc. once they are ready to be implemented. 

Requirements
===
use semantic dropdown option combined with dropdown.menu to create the menu items. 

Here are some examples:
https://github.com/Semantic-Org/Semantic-UI-React/blob/master/docs/app/Layouts/FixedMenuLayout.js
https://react.semantic-ui.com/modules/dropdown

PR #81, opened 2018-02-19
80 WIP update menu with better dropdown for logged in users.

mikelax:
This is very much a WIP. @nazar  there are a few issues that I would like you to take a look at. 

- [x] If use is logged out, then logs in, the process hangs on the /callback page. If you then refresh the page things load correctly. It must be something related to the meQuery.
- [x] There is a TODO about needing to use lodash to get the username as the component is rendering several times and the first couple the `me` object is null.
- [x] I can't figure out how to get the profile image to display next to the username in the menu, as opposed to inside the dropdown. 
- [x] I tried changing the Menu.Item objects to Dropdown.Item but that was causing issues and they weren't displaying correctly (couldn't see the text. 
- [x] If you click an option in the Dropdown.Menu you are taken to the new page, but the menu doesn't collapse. Similarly if the user clicks the username or "down arrow" that will activate the menu and keep the menu open until they click elsewhere on the page to close it.

### DIFF
--- a/client/src/components/Header/index.js
+++ b/client/src/components/Header/index.js
@@ -2,7 +2,6 @@ import _ from 'lodash';
 import React, { Component } from "react";
 import { Link, NavLink } from 'react-router-dom';
 import { connect } from 'react-redux';
-import { compose } from 'recompose';
 
 import logo from './logo.png';
 import './Header.styl';
@@ -138,11 +137,9 @@ const LeftMenuItems = connect(
   mapStateToProps, null, null, { pure: false }
 )(LeftMenuItemsBase);
 
-const RightMenuItems = compose(
-  connect(
-    mapStateToProps,
-    mapDispatchToProps, null, { pure: false }
-  )
+const RightMenuItems = connect(
+  mapStateToProps,
+  mapDispatchToProps, null, { pure: false }
 )(RightMenuItemsBase);
 
 export default class Header extends Component {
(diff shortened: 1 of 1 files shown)


client/src/components/Header/index.js
```diff
@@ -116,14 +111,27 @@ class RightMenuItemsBase extends Component {
       </React.Fragment>
     );
   }
+
+  _userLogo() {
+    const { me: { me } } = this.props;
+    const userProfileImageUrl = _.get(me, 'profileImage.url');
```
mikelax: @nazar  I guess we should just assume the pattern should be to use the lodash `get` method to get props from objects returned from redux / apollo ?

ie. when I do a fresh pageload already logged in, this console log returns undefined three times and then a fourth displays the actual value. 

```js
console.log(`the username is: ${username}`);
```

nazar: Hey @mikelax - that is correct and normal behaviour primarily because both [redux](https://github.com/mikelax/forgingadventures/blob/master/client/src/reducers/me.js) and apollo based object go through various states: i.e. loading, loaded (this has data) and error. Each of this stages triggers a re-render. 

If you add a console.log on the props object (use JSON.stringify as chrome console always displays the latest object values) - i.e. `console.log('props', JSON.stringify(this.props))`

If we don't care about those states then using `_.get(...)` is fine as otherwise we'll have to use something like:

```
if (me.loading) {
  foo
} else if (me.error) {
  foobar
} else {
  ...
}
```

nazar (approved):
:120%: 👍  :shipit:

mikelax:
Closes #80

Issue #38, opened 2018-01-13
Implement Quote Game (or Lounge) Message


mikelax:
Overview
===
As a user, I want the ability to be able to quote an already created Game Message. 

Requirements
===
There should be an option the user can select to "Quote" a given message. It should be displayed in a similar UI fashion as the Edit functionality. 
If quoted, then the editor should display the given message in a "Quoted" state. The user will then be able to add additional text before or after the quote block. 

The UI for displaying a quote can be a "light blue" solid background. There should be a header for the section that says who (or which character) wrote the original quoted text. 

There should also be a "Quote" menu bar button in the message editor. When clicked it should add the "Quote" block where the user can then add text inside the quote block. 

Questions
===
1. What do we do about nested quoting? ie. quoting a quoted message?

Issue #44, opened 2018-01-17
Bug with Message editor and "styled" text in last line


mikelax:
Overview
===
There is currently a bug in the editor when the last line of text uses one of the style icons. In this use case, if the cursor is after this style block (ie. the very end of the message) and you hit the return key, a newline is inserted before the style block instead of after it. 

This happens on both the initial message create, and also if you go back and edit a message.

Here is a sample message that shows the error use case. 

```
Here is the first line of text.
singingHere is a singing line of textsinging
```

This same issue also happens with the new Quote feature. It appears to be related to block level items. 
[This plugin](https://github.com/icelab/draft-js-block-breakout-plugin) appears like it might provide the fix we need. The plugin requires [this plugin framework](https://www.draft-js-plugins.com/) for draftjs. The framework might be a nice win anyway as there are some other plugins we may want to use such as emoji, image, linkify among others (these plugins will be in a separate Issue).

mikelax:
draftjs specific editor.

PR #82, opened 2018-02-19
Implement Quote Game (or Lounge) Message

nazar:
Refs #38


nazar:
Hey @mikelax 

The quoting feature isn't perfect but hopefully close enough. Issues include:

* not being able to nest quotes - we'll need to implement our own custom quote component to implement this
* the same message can't be quoted more than once via the Quote link. i.e. quoting a message twice via the Quote link only adds it a single time to the current editor. This one is slightly trickier - I need to dive deeper into DraftJS, hopefully after we get more features out of the way.

mikelax:
thanks @nazar , this is looking good. For your two points in your comment above, bullet 1 is a nice to have, but not needed for MVP. I don't think that we need to worry about bullet 2, seems like a real edge case. 

After testing this PR, here are a few comments and questions. 

- [ ] Is there a way we can add a small header on the top of the quote section. The ideal would be to contain text like:  Previous Message from :Name. To make it easier for the MVP we could omit the name and just have text say:  Previous Message. The header would have a slightly darker background color as the quote section color. 
- [ ] Is there a way you can insert a newline after the quote section? Currently it is hard to add more text after you have quoted text. 
- [x] The format buttons don't appear to work on the Game lounge messages. 
- [ ] In the game message editor, there isn't a way to insert a newline *after* the singing format. The shout and think formats work, so I suspect this error is due to the "sing icons" added to the editor. I think this is already reported in #44 


I know only some of these are actually related to this specific PR, so let me know if you want me to create a separate ticket for any of these items.

nazar:
Thanks for the feed back @mikelax - apart from the format buttons fix the other points are non trivial - i.e. it would require a custom quote component which would take about 1+ week to implement based on current velocity.

mikelax:
@nazar ok. Do you think it makes sense to  combine this bullet " Is there a way you can insert a newline after the quote section? Currently it is hard to add more text after you have quoted text." into the existing #44 which already covers the last bullet? 

I can then create a separate Issue for the first bullet about the header for the quote object?

nazar:
Hey @mikelax 

> Is there a way you can insert a newline after the quote section?

I had a look at this and there isn't a trivial way of adding a blank line after inserting content. The code in [this plugin](https://github.com/icelab/draft-js-block-breakout-plugin/blob/master/src/index.js) might be similar to what we want to do... tis not trivial.

mikelax:
@nazar  based on that README that does sound like what we want. 
Should we look to install and configure the [plugins framework](https://www.draft-js-plugins.com/) so we can make use of this plugin, along with others. ie. emoiji, image, linkify, etc?

mikelax (approved):
Created #87 and updated #44, so this PR is good to go!

nazar:
>Should we look to install and configure the plugins framework so we can make use of this plugin,

This plugin doesn't solve our issue but it has similar logic that __might__ be useful in our case - i.e. adding a line break at the end of a block. The plugin itself will not help us as it is focused at specific blocks (i.e. headers) and relates to keyboard based input and not programatically adding content to the editor as we are attempting.

The other plugins (i.e. emoji) would definitely be nice as does the image plugin (with it's alignments menu); we would have to factor in our own image upload logic and will most probably need a modal form where the user can either supply an image link (maybe dangerous?) or upload an image for us to process in cloudinary before we insert the image block into the editor.

mikelax:
@nazar what if we use the plugin as it is intended. The reason I had asked to insert an extra line after the quote was because as it stands now if you add a quote, then hit enter you are still in the quote section. There is no way to break out and end it. 

Isn’t that what this plugin does?  I will look at a couple of other editors to see how they allow you to add a return to a block element but then also end the block. 
Makes re the image upload logic.

nazar:
>what if we use the plugin as it is intended. The reason I had asked to insert an extra line after the quote was because as it stands now if you add a quote, then hit enter you are still in the quote section. There is no way to break out and end it.

Oooo! That would work!! I.e. the user can break out of it if they hit enter/return 👍 - that would make it much easier!

Unfortunately this plugin will only work on the built-in draftjs blocks (i.e. headers, blockqutoes, etc) and not any custom blocks/styles we add (i.e. singing, shouting, etc) as retaining the current style until it is toggled off is the default daftjs behavior - for example select H1 [here](https://draftjs.org/) and it remains until it is toggled off.

Issue #78, opened 2018-02-17
Can't sign up with new user in master (after #63)


mikelax:
Overview
===
In testing a previous ticket I had to do a docker down, which totally reset my db so I had to bring everything up from scratch. This included creating new users. 

While running master, just after #63 was merged, I am no longer able to sign up with a new email/password. 

The user is created within Auth0
The user is redirected to the /login/callback page. 
The various attributes are being written to the local storage correctly.
The page briefly says "Logging In" and then that text disappears. 

There is an error in the console as follows. I think the user is never actually getting inserted into the users table, so on the callback page when it tries to reference `me` that query has failed as the new user has not been inserted into the db at this point. I can't figure out where the "correct" place is to do this insert (or where it used to happen before this refactor. 

```
api_1     | 2018-02-17T03:24:34.866Z - info: select "users".* from "users" where "auth0UserId" = $1 - [auth0|5a8799bb5c679b17810f9c01] - 2.000 ms
api_1     | GraphQLError: User not found
api_1     |     at Object.__WEBPACK_IMPORTED_MODULE_2_models_user__.a.query.where.first.then.dbResult (/app/dist/webpack:/services/user.js:18:1)
api_1     |     at runCallback (timers.js:789:20)
api_1     |     at tryOnImmediate (timers.js:751:5)
api_1     |     at processImmediate [as _immediateCallback] (timers.js:722:5)
api_1     |     at QueryBuilder.then (/app/node_modules/objection/lib/queryBuilder/QueryBuilder.js:414:25)
api_1     |     at __WEBPACK_IMPORTED_MODULE_1_bluebird___default.a.try (/app/dist/webpack:/services/user.js:12:1)
api_1     |     at getUser (/app/dist/webpack:/services/user.js:11:1)
api_1     |     at runIfContextHasUser (/app/dist/webpack:/services/user.js:91:1)
api_1     |     at me (/app/dist/webpack:/schemas/user.js:40:1)
api_1     |     at resolveFieldValueOrError (/app/node_modules/graphql/execution/execute.js:498:12)
api_1     |     at resolveField (/app/node_modules/graphql/execution/execute.js:462:16)
api_1     |     at /app/node_modules/graphql/execution/execute.js:311:18
api_1     |     at Array.reduce (<anonymous>)
api_1     |     at executeFields (/app/node_modules/graphql/execution/execute.js:308:42)
api_1     |     at executeOperation (/app/node_modules/graphql/execution/execute.js:236:122)
[...]

PR #83, opened 2018-02-19
Can't sign up with new user in master

### LINKED ISSUE #63: Cleanup Client login logic

nazar:
closes #63

### DIFF
--- a/api/services/user.js
+++ b/api/services/user.js
@@ -88,7 +88,7 @@ export function runIfContextHasUser(context, runFn) {
   const token = _.get(context, 'req.user.sub');
 
   if (token) {
-    return getUser(token)
+    return getOrCreateUserByAuth0Id(token)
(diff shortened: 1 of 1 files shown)


mikelax (approved):
💥

mikelax:
Closes #78

Issue #67, opened 2018-02-07
Add ability to edit a Game


nazar:
Overview
===
After a game has been created, the GM (or game owner if we are using that status) should be able to update the game details. 

Requirements
===

* Extract the create game form into a component which can then be re-used in both the create game and edit game views.
* For this ticket, we can create the new page and routes for the "edit game" page, but do not need to create any direct links to it yet. We will do that once we implement a "My Games" section under the user profile. 
* See the attached page mock, we will implement a basic tab menu to allow the user to edit the game details, player status, all from a single location. 
  * The existing Players tab should be refactored into a component so it also can be used in two places, the existing Games page, and this new edit game page. 

Page mockup
---
![img_0003](https://user-images.githubusercontent.com/2080701/36643275-1f6fef84-1a17-11e8-8877-0cf5250cd9de.JPG)

PR #89, opened 2018-02-25
67 WIP for supporting edit game.

mikelax:
WIP, but ready for a quick review. 
Items still to do:

- [x] Refactor CreateGame into component for create & save
  - [x] Refactor graphql `gameQuery` variables to not be hard-coded value for edit, and to work for create
  - [x] Refactor `componentWillReceiveProps` function to work on both edit and create
  - [x] Refactor how submit action is handled so save action works for edit
  - [ ] Make Game Label read-only for edit
- [x] Go to edit form, cancel then use browser back button, get blank form instead of the edit version, also an issue with blank form when switching tabs on edit page
- [x] Add Auth so only GM can get to /edit
- [ ] Change URL when changing tabs
- [x] Have labels on the "Players" tab display correct number based on the data for each

Bonus
- [ ] Add standard browser save/cancel if user tries to navigate away from edit page before saving


client/src/components/Games/Create/index.js
```diff
@@ -40,124 +14,9 @@ class CreateGame extends Component {
...
-              />
-            </Form.Field>
-
-            <div className="actions text-right">
-              <Button primary onClick={this._submit} disabled={this.state.saving} loading={this.state.saving}>Submit</Button>
-              <Button onClick={this._cancel}>Cancel</Button>
-            </div>
-          </Form>
-
+          <GameDetailsForm history={this.props.history} cancelFn={this._cancel} />
```
mikelax: @nazar  is it correct to have to pass the `history` object down as a prop here?

nazar: Hey @mikelax 

The recommended way is to inject the history via react-router's [WithRouter](https://reacttraining.com/react-router/web/api/withRouter)

mikelax: Ah thanks. I saw the docs and *thought* that might be the right way. Will update this am.

mikelax: @nazar I will have to use the quick solution from this page to ensure re renders work since we are using redux, right?

https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/blocked-updates.md

nazar: Hmm - not sure about that - afaict, the GameDetailsForm doesn't use redux so we don't need the `connect` method.

Afaict, that is only needed in some situations where the redux `connect` function blocks non-specified props, which is what I saw happening on the navbar and had to fix it [here](https://github.com/mikelax/forgingadventures/blob/master/client/src/components/Header/index.js#L145) by specifying the `{pure: false}` function - but that is only because the Header component directly used redux store [values](https://github.com/mikelax/forgingadventures/blob/master/client/src/components/Header/index.js#L135-L136).

client/src/components/Games/components/GameDetailsForm/index.js
```diff
@@ -0,0 +1,259 @@
...
+
+          if (gameImage) {
+            payload.gameImage = {
+              publicId: _.get(gameImage, 'publicId'),
+              userUploadId: _.get(gameImage, 'userUploadId'),
+              url: _.get(gameImage, 'imageUrl')
+            };
+          }
+
+          this.setState({ saving: false });
```
mikelax, line 261: @nazar  I had to move this setState here above the save function call as otherwise i was getting a browser warning for:  
```
Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the undefined component.
```

mikelax:
@nazar  do you have any insights on this item? `Go to edit form, cancel then use browser back button, get blank form instead of the edit version, also an issue with blank form when switching tabs on edit page`
It must have something to do with the component lifecycle, as when i am looking at the page and switching tabs i can see the props sent to the component look correct, so somewhere the state is getting lost.

mikelax:
@nazar  nm RE my last comment, I fixed the issue by implementing the `componentWillMount` function.

mikelax:
Hey @nazar  question I wrote the new guard for viewing the game edit, its almost working, but since it must trigger several renders because of the graphql addition, it fails before all data is loaded,  ie. when one var is 13 and the other is undefined. 

see screenshot of debug console.logs.
any thoughts on how to get or defer the check until all data is loaded (or a better way to load data)?
screen shot 2018-02-28 at 10 44 08 pm


PS. I was also thinking, is this guard overkill for what we are trying to protect? Should I have just done a simple check in the `render` method and performed a redirect there?

nazar:
Hey @mikelax ,

>see screenshot of debug console.logs.
any thoughts on how to get or defer the check until all data is loaded (or a better way to load data)?

This is the standard behaviour of React and is something that should be taken as a given. The key selling point of React is that generally the developer doesn't care how many times the component's render method is called (as long as it isn't expensive but generally these are super cheap) as it is React's diffing engine that is finally responsible for altering the DOM (altering the DOM is expensive)

This means that we no longer need to write optimised code - or code that syncs different async operations as the diffing engine takes care of that by batching all these renders into smaller or a single chunk. So in theory we just concentrate on business logic and not worry about difficult/messy code that generally results out of optimisation.

This, on the other hand, means that whenever we access props, we cannot guarantee how many times a render method will be called and whether props have data, especially if say a component is connected to multiple graphQL queries and/or multiple redux stores. Generally a change in any of those triggers a render - the more connected stores the more times the render method is called for each change.

So we should always a assume that a prop we need will most probably start as undefined and then in subsequent renders/licefycle updates have a value. So pattens such as `_.get(props, 'foo.bar)` or the [upcoming optional chaining notation](https://dev.to/sammyisa/optional-chaining-may-be-coming-to-javascript--4ff0) should be used if we want to access values in props.

>PS. I was also thinking, is this guard overkill for what we are trying to protect? Should I have just done a simple check in the render method and performed a redirect there?

We could also do that but having the owner guard is a nice way of code re-use, especially if we need to re-use that same logic in  other parts of the code. If the owner guard is only applicable to editing a game then it could be argued that it is OTT. I've just pushed an example implementation of the owner guard.

HTH

mikelax:
thanks @nazar, really helpful info. 
So I refactored that code snippet a bit to wait until all props are loaded, does this type of pattern make sense? 

```js
        // wait until all props are loaded
        if (_.isUndefined(userId) || _.isUndefined(gameOwnerId)) {
          return true;
        } else {
          return _.eq(userId, gameOwnerId);  
        }
```

nazar:
```js
        // wait until all props are loaded
        if (_.isUndefined(userId) || _.isUndefined(gameOwnerId)) {
          return true;
        } else {
          return _.eq(userId, gameOwnerId);  
        }
```

I think we want the opposite? i.e. return `true` only if `userId` and `userId` and not falsy and are not equal to render the [left branch](https://github.com/acdlite/recompose/blob/master/docs/API.md#branch).

client/src/components/Games/Edit/index.js
```diff
@@ -0,0 +1,124 @@
...
+
+import OwnerGuard from '../../shared/components/OwnerGuard';
+
+class EditGame extends Component {
+  state = { activeItem: 'details' };
+
+  render() {
+    const { activeItem } = this.state;
+    const title = _.get(this.props, 'data.game.title');
+    const playerCounts = _.groupBy(_.get(this.props, 'gamePlayers.gamePlayers'), 'status');
```
nazar, line 20: 👍 ❤️ 💯

mikelax:
Closes #67

PR #93, opened 2018-03-04
90 profile page

### LINKED ISSUE #90: Convert Profile Page to User Profile Page
Overview
===
As a user, I want to view my profile page, which will display some of my basic profile fields like name, image, username, and then have tabs for my games and characters. 

Requirements
===

* This page will be different from a "settings" page. That page will allow the user to update their profile data, change profile image, etc. 
* The URL for the page will be:  /profile
 * In the future we will change this URL to be:  /:username
* There will be a tab bar section that will contain tabs for Games and Characters (once implemented)

Page mock
---
profile-page

mikelax:
Closes #90. 
WIP, basic page structure is in place with queries to get the user object and games list. 
Still a few questions to be answered, depending the work may be done in this branch or as separate tickets. 

- [ ] Should we have a new page that will serve as the "public view" of a user, ie. the real profile page, and this page should turn into an "/account" page? Or is there too much overlap and we should have a single profile page and just change some options available depending on the user viewing the page (ie. owner vs. regular user)? 
- [x] How should we get the list of pending games? I can see a few different ways of but wanted to get your opinion on modifying / creating new `Game` type vs. `GamePlayer`. 
  - [ ] ie. We may also want to do something similar so we can indicate if the user is a GM for any of the games in the Games list. For v1 maybe we just look at the "userId" attribute


mikelax:
We can add a label using something like the following to get an "Edit" icon for games the user is a GM for. 

```jsx
<Label attached='bottom right' icon='setting'></Label>
```

mikelax:
@nazar  thanks for the refactor, the new profile updates look good. Couple of questions:
1. More a UX question. Do you think we should display all the extra tabs like kicked, rejected, and Quit?  Or maybe group them all together into one single tab?
1. In a future ticket should we refactor this page to separate "completed" games, or active games where the user is a "GM" vs. "player"?
  1. Ie. this gets into a two variable matrix (probably way too complicated for what we are trying to do). ie. User is GM and Game is Active, User is Player and Game is Active, User is Player and Game is Completed, etc.

nazar:
>More a UX question. Do you think we should display all the extra tabs like kicked, rejected, and Quit? Or maybe group them all together into one single tab?

Good question - ultimately the groups makes more sense but maybe for a v2?

>In a future ticket should we refactor this page to separate "completed" games, or active games where the user is a "GM" vs. "player"?

Yeah I did think about that as well and tried to do a quick GMing/Playing split but couldn't get it quite right. I thought I best leave it for v2

>this gets into a two variable matrix (probably way too complicated for what we are trying to do). ie. User is GM and Game is Active, User is Player and Game is Active, User is Player and Game is Completed, etc.

yep yep - V2 :D

mikelax:
Merging PR, remaining two tasks will be pushed to separate tickets to implement.

PR #95, opened 2018-03-05
94 WIP for game status support.

### LINKED ISSUE #94: Implement Game Status, including with Search and Game Edit
Overview
===
As a GM for a game, I need to be able to update the status for a game. ie. Change from looking for players to running, or any other status available. 
As a regular user, I want to be able to search for a Game on the Games Search page by Game status, among other search options. 

Requirements
===
In the `gameSettings.js` file add additional values to the `gameStatus` attribute. 
4 - Completed

Create Game
---
After a new game is created, it should receive an initial game status of:  1:  Recruiting Players. 
This will allow it be immediately be discoverable on the search page. 

Edit Game
---
As the GM of a game, in the Edit Game screen, the GM will be able to change the game status. 
There should be a new tab added to the Edit screen called "Status". It should be added after the "Players" tab. 

This tab should display the current status. It should then allow the GM to change the status to another value. 
TODO:  should this just be a boring dropdown, or something with a bit more flair, maybe buttons to showing the status "moving".

Search Games
---
The status attribute in the search form should be updated to be a multi-select instead of a dropdown. 
When the page loads the default values selected for the status search are:  1,3

mikelax:
Closes #94 
WIP for full support of game status. 
Items still TODO: 

- [x] Implement updating status on edit game page
- [ ] MAYBE:  convert status dropdown on game search page to multi select

### DIFF
--- a/api/schemas/game.js
+++ b/api/schemas/game.js
@@ -14,7 +14,8 @@ export const gameTypeDefs = `
     minPlayers: Int!,
     maxPlayers: Int!,
     skillLevel: Int!,
-    postingFrequency: Int!
+    postingFrequency: Int!,
+    gameStatus: Int!
   }
   
   type Game {
@@ -84,7 +85,10 @@ export const gameResolvers = {
             return Game
               .query()
               .insert(_.merge({}, {
-                userId: user.id
+                userId: user.id,
+                gameSettings: {
+                  gameStatus: 1
+                }
               }, input))
               .returning('*');
           })
--- a/api/schemas/game.js
+++ b/api/schemas/game.js
@@ -5,7 +5,7 @@ import GamePlayer from 'models/gamePlayer';
 
 import schemaScopeGate from 'services/schemaScopeGate';
 import GetGames from 'services/games/getGames';
-import { getUser } from 'services/user';
+import { getUser, runIfContextHasUser } from 'services/user';
 
 import serviceExecutor from 'utils/serviceExecutor';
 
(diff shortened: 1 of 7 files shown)


mikelax:
@nazar  this PR will require any existing games to have a `gameStatus` attribute added to the `gameSettings` column.

api/schemas/game.js
```diff
@@ -84,7 +85,10 @@ export const gameResolvers = {
             return Game
               .query()
               .insert(_.merge({}, {
-                userId: user.id
+                userId: user.id,
+                gameSettings: {
+                  gameStatus: 1
```
mikelax, line 90: @nazar  here is where i am adding the default game status of "Looking for Players"

client/src/components/Games/Edit/index.js
```diff
@@ -108,9 +114,68 @@ class EditGame extends Component {
...
+              Update Game Status
+            </Button>
+          </Form>
+      </React.Fragment>
+    );
+  };
+
+  _submit = () => {
+    const { onSave } = this.props;
+    const payload = _.merge({}, this.props.game);
```
mikelax: @nazar  I saw this pattern used in the other submit. 
I tried something like this but got an error about not being able to set a readOnly attribute `gameStatus` in the following line. 

```js
const { game } = this.props;
game.gameSettings.gameStatus = Number(this.state.status);
```

mikelax:
Merging PR for now. we may come back to how the game mutations are handled in a future ticket to reduce similar/copied code to verify user ownership as first step.

PR #99, opened 2018-03-10
#97 adds player location to Game players tab.

### LINKED ISSUE #97: Add location to Players Tab in Game View Page
Overview
===
As a user, I would like to be able to see the locations of the other Players (and GM) for a given Game. 

Requirements
===
On the Players tab in Game View, add a new field for the Player Location. 
At a minimum, it should display the Location Value, ie:  `America/New_York`, a bonus would be if we could also display the timezone offset. ie. 
`America/New_York (UTC-4)`, 

maybe even also display the Player's local time?

mikelax:
Closes #97

### DIFF
--- a/api/models/gamePlayer.js
+++ b/api/models/gamePlayer.js
@@ -6,15 +6,4 @@ const mixins = compose(setUpdatedAt);
 
 export default class GamePlayer extends mixins(Model) {
   static tableName = 'game_players';
-
-  static modelPaths = [__dirname];
-
-  static relationMappings = {
-    relation: Model.BelongsToOneRelation,
-    modelClass: 'User',
-    join: {
-      from: 'game_players.userId',
-      to: 'users.id'
-    }
-  }
 }
--- a/api/models/gamePlayer.js
+++ b/api/models/gamePlayer.js
@@ -6,4 +6,15 @@ const mixins = compose(setUpdatedAt);
 
 export default class GamePlayer extends mixins(Model) {
   static tableName = 'game_players';
+
+  static modelPaths = [__dirname];
+
+  static relationMappings = {
+    relation: Model.BelongsToOneRelation,
+    modelClass: 'User',
+    join: {
+      from: 'game_players.userId',
+      to: 'users.id'
+    }
+  }
 }
--- a/client/src/components/Games/View/components/GamePlayers.js
+++ b/client/src/components/Games/View/components/GamePlayers.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { graphql } from 'react-apollo';
 import { compose, pure } from 'recompose';
-import { Button, Header, Image, List, Menu, Popup, Icon, Table } from 'semantic-ui-react';
+import { Button, Header, Image, Menu, Popup, Icon, Table } from 'semantic-ui-react';
 
 import { gamePlayersQuery, myGamePlayerQuery, updateGamePlayerMutation } from '../../queries';
 import ApolloLoader from '../../../shared/components/ApolloLoader';
--- a/client/src/components/Games/View/components/GamePlayers.js
+++ b/client/src/components/Games/View/components/GamePlayers.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { graphql } from 'react-apollo';
 import { compose, pure } from 'recompose';
-import { Button, Image, List, Menu, Popup, Icon } from 'semantic-ui-react';
+import { Button, Header, Image, List, Menu, Popup, Icon, Table } from 'semantic-ui-react';
 
 import { gamePlayersQuery, myGamePlayerQuery, updateGamePlayerMutation } from '../../queries';
 import ApolloLoader from '../../../shared/components/ApolloLoader';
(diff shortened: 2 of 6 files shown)


api/models/gamePlayer.js
```diff
@@ -6,4 +6,15 @@ const mixins = compose(setUpdatedAt);
 
 export default class GamePlayer extends mixins(Model) {
   static tableName = 'game_players';
+
+  static modelPaths = [__dirname];
+
+  static relationMappings = {
```
mikelax: @nazar  do you think we should look to use the objection relationships, instead of using the apollo resolvers which I assume has to make extra db hits to fetch the additional data.

nazar: >do you think we should look to use the objection relationships, instead of using the apollo resolvers which I assume has to make extra db hits to fetch the additional data.

Hey @mikelax - I'm more tempted to stick with apollo resolvers as much as we can as that forces us to describe our schema as a graph and keeps all relationships defined in the graphQL namespace.

In terms of extra db hits, I think that's fine as imho smaller queries (i.e. ID PK based lookups) scale better than complex join queries especially once we start applying [dataLoader](https://github.com/facebook/dataloader) on our relations.

I'm sure we won't be able to stick 100% to graphQL relationships but I definitely think we should minimise joins, especially when querying data (not so much when we update). We definitely should not use objection relationships to load associated data.

client/src/components/Games/View/components/GamePlayers.js
```diff
@@ -25,24 +25,39 @@ class GamePlayers extends Component {
...
+                  <Header as='h3' image>
+                    <Image avatar src={_.get(player, 'user.profileImage.url')} />
+                    <Header.Content>
+                        {player.user.name}
+                      <Header.Subheader>{player.user.timezone}</Header.Subheader>
+                    </Header.Content>
+                  </Header>
+                </Table.Cell>
+                <Table.Cell>
+                  { player.status === 'accepted' ? <Icon name='checkmark' /> : null }
```
nazar, line 49: that table looks MUCH nicer 👍

api/models/gamePlayer.js
nazar: Hey @mikelax ,

Creating a game was throwing an error for me on the `gamePlayer` model - is that only on my side or is it happening for you as well?

```
api_1     | 2018-03-12T14:12:55.699Z - info: insert into "games" ("gameSettings", "labelId", "overview", "scenario", "title", "userId") values ($1, $2, $3, $4, $5, $6) returning * - [{"gameStatus":1,"minPlayers":2,"maxPlayers":6,"skillLevel":1,"postingFrequency":1},2,gg,gg,ggggg,3] - 3.000 ms
api_1     | GraphQLError: mapping.relation is not a constructor
api_1     |     at Object.keys.reduce (/app/node_modules/objection/lib/model/Model.js:423:35)
api_1     |     at Array.reduce (<anonymous>)
api_1     |     at Function.getRelations (/app/node_modules/objection/lib/model/Model.js:420:61)
api_1     |     at getRelationArray (/app/node_modules/objection/lib/model/Model.js:652:32)
api_1     |     at Function.getRelationArray (/app/node_modules/objection/lib/model/Model.js:439:60)
api_1     |     at parseRelationsIntoModelInstances (/app/node_modules/objection/lib/model/modelParseRelations.js:5:32)
api_1     |     at GamePlayer.$setJson (/app/node_modules/objection/lib/model/modelSet.js:23:5)
api_1     |     at GamePlayer.$setJson (/app/node_modules/objection/lib/model/Model.js:163:21)
api_1     |     at Function.fromJson (/app/node_modules/objection/lib/model/Model.js:282:11)
api_1     |     at doSplit (/app/node_modules/objection/lib/model/modelFactory.js:141:22)
api_1     |     at obj.map.obj (/app/node_modules/objection/lib/model/modelFactory.js:74:28)
api_1     |     at Array.map (<anonymous>)
api_1     |     at fromJsonShallow (/app/node_modules/objection/lib/model/modelFactory.js:74:17)
api_1     |     at fromJson (/app/node_modules/objection/lib/model/modelFactory.js:14:12)
api_1     |     at InsertOperation.onAdd (/app/node_modules/objection/lib/queryBuilder/operations/InsertOperation.js:38:21)
api_1     |     at QueryBuilder.addOperation (/app/node_modules/objection/lib/queryBuilder/QueryBuilderOperationSupport.js:147:33)
api_1     |     at QueryBuilder.execute (/app/node_modules/objection/lib/queryBuilder/QueryBuilder.js:527:20)
api_1     |     at QueryBuilder.then (/app/node_modules/objection/lib/queryBuilder/QueryBuilder.js:413:26)
api_1     |     at runCallback (timers.js:789:20)
api_1     |     at tryOnImmediate (timers.js:751:5)
api_1     |     at processImmediate [as _immediateCallback] (timers.js:722:5)
api_1     |     at Object (/app/dist/webpack:/schemas/game.js:104:1)
api_1     |     at __webpack_exports__.a (/app/dist/webpack:/services/schemaScopeGate.js:25:1)
api_1     |     at createGame (/app/dist/webpack:/schemas/game.js:90:1)
api_1     |     at resolveFieldValueOrError (/app/node_modules/graphql/execution/execute.js:498:12)
api_1     |     at resolveField (/app/node_modules/graphql/execution/execute.js:462:16)
```

If I update the model to the following then I am able to create a game:

```
import { compose, Model } from 'objection';

import setUpdatedAt from 'models/plugins/setUpdatedAt';
import User from 'models/user';

const mixins = compose(setUpdatedAt);

export default class GamePlayer extends mixins(Model) {
  static tableName = 'game_players';

  static relationMappings = {
    user: {
      relation: Model.BelongsToOneRelation,
      modelClass: User,
      join: {
        from: 'game_players.userId',
        to: 'users.id'
      }
    }
  }
}

```

nazar: >do you think we should look to use the objection relationships, instead of using the apollo resolvers which I assume has to make extra db hits to fetch the additional data.

Follow up - 

Looking at the docs, defining relations might give us a [nice way of manipulating data](http://vincit.github.io/objection.js/#insert-queries13). There is definitely value in also [fetching](http://vincit.github.io/objection.js/#fetch-queries12) if we want to query data in resolvers.

I don't think we can use it to return relational data for consumption by the client via graphQL Queries, however, mainly due to how graphQL schemas work.

In a schema, graphql will traverse the tree if a non scaller field is encountered (i.e. `User` is not scaller but `String` is). It is this traversal that allows us to construct queries in the client that can go n levels deep.

I strongly suspect we would lose this n-depth query ability if objection were to take care of the association queries.

mikelax:
@nazar  I removed the `objection.js` relations for now.
We can revisit in a separate ticket if we decide to actually implement them.

PR #112, opened 2018-03-22
#109 WIP edit characters

### LINKED ISSUE #109: Add edit and status (in game / available) to Character Card on User Profile page
Overview
===
As a User, I want to be able to edit an existing Character I have already created. While looking at the Characters tab on the User Profile, I should be able to click a "edit" icon to get to the edit screen. 

The Character card should also display the status of a Character. It should say if it is available, or currently in a Game. If in a game, that should be a link to jump to that Game. 

Requirements
===

* Add a new Settings icon to the Card, probably the lower right corner. This should link to `/characters/:id/edit`
* Add a meta section to the card that displays the Game status. If the Character is in a game, that should be a link to the Games page: `/games/:id`

mikelax:
Closes #109 
Still a WIP, @nazar need to figure out how to write the resolver in the `character.js` schema file. I don't think we have had a situation yet where we need to connect two objects. 
ie. character -> game player -> game

Once this is resolved (haha) the commented out code in the client `queries.js` file can be added to the `myCharactersQuery` query.

### DIFF
--- a/api/schemas/character.js
+++ b/api/schemas/character.js
@@ -11,6 +11,7 @@ export const characterTypeDefs = `
     profileImage: ProfileImage,
     user: User!,
     label: GameLabel!,
+    gamePlayer: GamePlayer,
     updated_at: GraphQLDateTime,
     created_at: GraphQLDateTime
   }
@@ -30,7 +31,12 @@ export const characterTypeDefs = `
 
 export const characterResolvers = {
   Character: {
-    label: (character, vars, context) => context.loaders.gameLabels.load(character.labelId)
+    label: (character, vars, context) => context.loaders.gameLabels.load(character.labelId),
+    gamePlayer: (character) => {
+      return GamePlayer.query()
+        .whereIn('status', ['pending', 'accepted'])
+        .where({ characterId: character.id });
+    }
   },
   Query: {
     availableCharacters: (obj, { gameId }, context) => {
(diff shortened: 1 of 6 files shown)


nazar:
`Once this is resolved (haha)` - 😂 got me there!

I'll have a look at this - in theory we should be able to go n levels deep (once we figure out how)

api/schemas/character.js
```diff
@@ -30,7 +31,12 @@ export const characterTypeDefs = `
 
 export const characterResolvers = {
   Character: {
-    label: (character, vars, context) => context.loaders.gameLabels.load(character.labelId)
+    label: (character, vars, context) => context.loaders.gameLabels.load(character.labelId),
+    gamePlayer: (character) => {
+      return GamePlayer.query()
+        .whereIn('status', ['pending', 'accepted'])
```
nazar: I don't think we can set the `status` condition here as the resolvers should be able to navigate the object graph without any restrictions. 

This filtering should be done upstream and `gamePlayer` should be considered as a zero or many relationship - `gamePlayer: [GamePlayer]`

nazar:
Hey @mikelax - 

>think we have had a situation yet where we need to connect two objects.
ie. character -> game player -> game

Looking at the schemas, `character -> game player -> game` should already exist as [game player is connected to a game](https://github.com/mikelax/forgingadventures/blob/109-edit-characters/api/schemas/gamePlayer.js#L16)

nazar:
screen shot 2018-03-22 at 12 23 40

PR #113, opened 2018-03-22
closes #104 - select character after joining

### LINKED ISSUE #104: Ability to select Character on Game -> Players tab
Overview
===
As a player that has already joined a Game, I want to be able to select a Character I have created already. This use case is where a User joined a Game but did not select a Character on the Join screen. This is the last task to implement from the #100 Issue. 

Requirements
===
If User is viewing a game they have joined, but have selected a Character, the Character column displays text:  "Still working on it". We should add a button or link to this column near (side or below) this text to link to the Create Character page. /characters/create

In the Game Actions column, a new button should be added above Leave Game with an option for "Select Character". This should bring up a modal / overlay to allow the user to select any existing Characters. It can use the newly created `CharactersSelect` component as this has the correct query already implemented.

nazar:


mikelax (approved):
🚀

PR #117, opened 2018-03-26
Refactor Game create/edit to use Rich text

### LINKED ISSUE #101: Update create/edit Game screen to use rich editor for "Overview" attribute and Scenario -> textfield
Overview
===
We should change the Overview field to use the new rich text editor instead of a basic textarea. This will allow the GM to create a nicer looking Overview and introduction for the players. 

Related to this change, we should change the `Scenario` field from a textarea to a regular textfield. This should be done for both the Game create and edit.

nazar:
For game overview and input for scenario

Closes #101


mikelax (approved):
:shipit:

Issue #114, opened 2018-03-24
Create (Joi) Schema and related code for v1 Character attributes


mikelax:
Overview
===
This ticket will cover the work to generate the schemas that will be used to store the attributes for a Character. 

Requirements
===

* The schema will contain the list of fields that will be displayed and used within the UI. 
* A new db field will be added to the characters table to store this attribute document, data type will be JSON, nullable. 
* We decided that separate Component Forms will be created that will be specific to each Game Label. 
  * The form should be able to handle some dropdown fields will a known set of values
* [Here is a simplified 5e character sheet](https://drive.google.com/file/d/1G7S5TF28i-kshGKNxsFLPbKACSc0bno_/view?usp=sharing) that we can use to model the schema from. Originally grabbed from [this blog post](http://skullboyrollsthedice.blogspot.com/2018/03/redesigning-5th-edition-character-sheet.html).
* [Here is a 5e character sheet](https://boardgamegeek.com/article/25994829#25994829) from BGG

Questions
---

1. Do we want to try to support dropdown values? 
1. Do we want to try to support generated attributes, ie. where a field could be altered by a value in a different attribute? 
  * Example of this is where Race can give certain attributes a bonus. 
1. Should pure formula values be included in JSON structure? For example "Ability Score Modifiers" is a pure formula based on the Ability Value. ie. Strength of 12 = Modifier of +1. Formula is: subtract 10 from ability, divide result by 2 (round down). (12-10)/2 = +1
  * Storing the values in the JSON might make certain other actions easier as you'll have the value available, rather than constantly needing to go and calculate. You only need to calculate on create/edit? Let me know your thoughts.

mikelax:
@nazar  here is what I came up with for a possible schema for 5e label. 
It's probably about 90% there, it might need some small changes around spells, etc. I need to research that more. 

Notes
1. The `meta` section is for fields used by the application. 
1. This json has fields that could be "calculated". ie. `raceBonus` is from a table that is a standard value based on Race. 
1. We have to decide what we want to do for values that would come from a dropdown, ie. for things like Race, Class, alignment, and a few more. We obv. know the potential issues with storing  the "human readable" value in the db, makes calculating logic with it more difficult. 

Let me know your thoughts, once we get answers to these couple of questions I think we can get the Joi schema going, add db field, mutations, etc. 

```
{
  "meta": {
    "schemaVersion": "1.0.0",
    "gameLabelId": 1
  },
  "primaryLevel": 1,
[...]

PR #119, opened 2018-03-27
#114 WIP for start of Joi schemas for character attributes.

mikelax:
@nazar  Here are a couple of starter Joi schemas. 
Would appropriate next steps be to: 
* Create migration to add nullable attributes column to characters table
* Refactor create / update character mutations to use this new column?


nazar:
Hey @mikelax - the Join schema looks 👍 

>Create migration to add nullable attributes column to characters table

We'll have to consider how normalised the `characters` columns should be. Should we have columns for `traits`, `health` etc or should the Joi schema describe a single JSONB column that contains all the character's details? A lot of that depends on how common base traits are and whether they work the same way across games: for example (a made up one) DnD health might be an INT but BnB (made up game) might have health that is broken into two fields.

I  am tempted to put everything in a single JSONB column but that feels a bit reckless to me!

>Refactor create / update character mutations to use this new column?

👍

mikelax:
Cool thx @nazar. 
I was thinking all the data from the joi schema would go into a single jsonb field. 

I think it will be too hard to try and match up fields between game systems. 

There *might* be some overlap with “d20” based systems, but there will be other games that are totally different. Ie maybe don’t even have health at all (for example).

nazar:
>I was thinking all the data from the joi schema would go into a single jsonb field.

👍 ❤️ 

>I think it will be too hard to try and match up fields between game systems

☝️

mikelax:
merging this PR. will do rest of work in #125 PR

PR #121, opened 2018-04-09
MISC Refactor docker files to support dev and production

mikelax:
run nginx for test/production.

This PR could be merged now, but additional tasks to get app working on ECS. 
- [x]  Need to add env variables when running on ECS.


client/package.json
```diff
@@ -32,8 +32,10 @@
     "recompose": "^0.26.0",
     "redux": "^3.7.2",
     "redux-thunk": "^2.2.0",
+    "rupture": "0.7.1",
     "semantic-ui-css": "^2.2.12",
-    "semantic-ui-react": "^0.78.2"
+    "semantic-ui-react": "^0.78.2",
+    "stylus-loader": "3.0.1"
   },
```
nazar: nitty: I'd keep both `stylus-loader` and maybe `rupture` in `devDependencies` as both are webpack and tooling related libraries.

nazar (approved):
Hey @mikelax - I just made a couple of tiny changes.

We don't need node in the final build so I swapped out `node:8.9.4-alpine` to `nginx:1.13-alpine` for an even smaller container.

client/package.json
mikelax: Hey @nazar the production build was failing because of missing Deps until I moved these two libs here. 

This was with the command yarn install —production

nazar: oh WOW! 👍  💯

nazar: now I'm wondering if we need `yarn install —production` - i.e. just use `yarn install`. Maybe `yarn install —production` makes more sense for a node application.

mikelax:
@nazar  why wouldn't we run install with `production`?
Also how come you removed the --cache? Is that the keep the layer from being invalidated, does that "undo" docker logic?

Issue #122, opened 2018-04-11
Upgrade "scaffolding" libs to newest version, starting with react 16.3.x


mikelax:
Overview
===
It has been some time since we last updated some of the major libs we are using in this project. 
We should upgrade to react 16.3.x, and make the other related package upgrades as well: 

* react router
* redux
* apollo / graphql

mikelax:
@nazar  FYI I'll take a first stab at this ticket.

mikelax:
After trying to upgrade to latest of apollo, related apollo libs, and graphql and graphql-tag, no worky. 
error when trying to load games page. 
```
TypeError: parse is not a function
parseDocument
node_modules/graphql-tag/src/index.js:128
  125 |   return docCache[cacheKey];
  126 | }
  127 | 
> 128 | var parsed = parse(doc);
  129 | if (!parsed || parsed.kind !== 'Document') {
  130 |   throw new Error('Not a valid GraphQL document.');
  131 | }
```

PR #124, opened 2018-04-12
122 WIP upgrade some minor packages.

mikelax:
sadface trying to upgrade apollo and graphql (tag). left comment in #122 with error, will have to try other lib combinations. 
Addresses #122


mikelax:
@nazar  I got all the apollo and react libs upgraded. 
So there is an [open ticket](https://github.com/apollographql/graphql-tag/issues/155) where graphql-tag does not work with graphql > 0.13.0

Also, I was able to just use the apollo-boost lib instead of bringing in all the libs separately. There are some peer warnings, its like a spiders web trying to get all the various lib versions to work together.

I think this is ready to go. Then I'll do a separate PR to upgrade the cra libs, and then one for the remaining dev deps like eslint, etc.

mikelax:
@nazar  ok I got the eslint upgrades done, and fixed a few errors in api. 
If you want a facepalm, here is the reason the jsx-a11y lib isn't updated in the client folder. 
TLDR, cra has issues using esllint, even after ejecting.... 😕 
https://github.com/facebook/create-react-app/issues/3418
https://github.com/odota/web/issues/1020#issuecomment-381165135
https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/397#issuecomment-362839806

This is ready to be merged though. 👍

nazar (approved):
:shipit:  👍  💯

PR #125, opened 2018-04-13
Character Label Forms

### LINKED ISSUE #115: Modify Character create/edit page to use label specific Form Component
Overview
===
As a User, when I create a new Character, I should be able to enter values for character attributes based on the _Game Label_ selected for the Character. 

Requirements
===
* On the create page, after the User selects a Label, the Form component for the given Label should be displayed. 
* Since the exact fields are still being defined in #114, this ticket can have a single textfield and custom text copy for each of the labels to show the process is picking up the correct form for a given Label. 
* We can limit the scope of this ticket to that of the all the UI components and logic. We'll do the work of updating mutations, new db column, etc. in a separate ticket. 

Nice to Have
---
* If the User selects a Label, enters some values into the Form, then changes the Label to a different value, the page should prompt the User with a confirm dialog asking if they are sure they want to change the Label and it will reset the Custom Attributes. 

Questions
===
1. After the initial create, should we not allow the User to change the _Game Label_ when editing a Character?
1. Should we make the create character a two step process, where the first step allows the user to select a label. Then for the second step we know the exact form to display? Otherwise on a single page, the "bottom half" of the form will get changed after the Game Label value is selected. 
  1. I _think_ I would prefer to keep the process to a single page, but open to your thoughts.

nazar:
Very very wip
Closes #115  and #114


client/src/components/Characters/components/CharacterLabelForms/CharacterDetails5e_1_0_0.js
```diff
@@ -0,0 +1,340 @@
...
+                <Form.Field required label="Current" name="health.currentHitPoints" control={Field} type="number"
+                            min="0"/>
+              </Form.Field>
+
+              <Form.Field required>
+                <Form.Field required label="Max" name="health.maxHitPoints" control={Field} type="number" min="1"/>
+              </Form.Field>
+
+              <Form.Field required>
+                <Form.Field required label="Hit Die" name="health.hitDie" control={Field} type="number" min="0"/>
```
mikelax: @nazar  this field needs to be a string as a hit die could be:  "1d10".

client/src/components/Characters/components/CharacterLabelForms/CharacterDetails5e_1_0_0.js
```diff
@@ -0,0 +1,340 @@
...
+        onSubmit={this._handleSubmit}
+        validationSchema={validationSchema}
+        render={({ values, handleChange, handleReset, handleSubmit, isSubmitting }) => (
+          <Form className="character-details" action={handleSubmit} loading={loading}>
+            <h2>Custom Character Settings</h2>
+
+            <Form.Group widths="equal">
+              <Form.Field required>
+                <label>Race</label>
+                <Field
```
mikelax: @nazar  I was just thinking that maybe we should use the react-select package for these dropdowns. 
ie. specificaly for race and class, we can have the dropdowns allow for custom text, as sometimes people user custom (homebrew) values for these. so the fields can auto-complete the common values, but allow free-text. only downside is that makes it hard to calculate modifiers based on race/class. maybe homebrew values are supported for a future version on second thought..... so this comment is really just FYI as a think while typing this out,

client/src/components/Characters/components/CharacterLabelForms/CharacterDetails5e_1_0_0.js
```diff
@@ -0,0 +1,358 @@
...
+  health: Yup.object().shape({
+    currentHitPoints: Yup.number().integer().default(0),
+    maxHitPoints: Yup.number().integer().default(1).min(1),
+    hitDie: Yup.string().default('')
+  }),
+  abilities: Yup.object().shape({
+    strength: Yup.object().shape({
+      baseValue: Yup.number().integer().default(1).min(1),
+      raceBonus: Yup.number().integer().default(0).min(0),
+      total: Yup.number().integer().default(1).min(1).max(20),
```
mikelax: @nazar  I added a `total` here. This is because as a character increases in Level, they are able to increase their ability scores. So for any given ability, the "total" number = base number + race bonus + level up increases

Maybe for a v1 it would be better to just store a total for each, we can then, in the future, implement the audit trail that will show how any given ability changed over time. 
Alternative would be to store:  baseValue, raceBonus, levelIncreases (this gets incremented as the user continues to increase levels), total. Thoughts?

mikelax:
@nazar  I was thinking about the attributes more and I think that we should use the fields that we have. 
ie, for example Strength:
Total (read only) = Base Value (editable) + Race Bonus (read only, or editable for mvp)
Modifier (read only) = (Total - 10 ) / 2   round down

As players level up, they can edit and increase the number in the `Base Value` attribute. 

For UI display, the Total should be the largest element, along with modifier. 
We can adjust the UI in a separate ticket once we get all fields and saving, etc working.

mikelax:
@nazar  thanks for the changes! 
for the dockerfile rename, instead of a move, do you want to do a copy, as I am currently using the existing dockerfile for the builds to AWS. Obv. we know they can be refactored to be closer to DS, but that can come in a separate PR.

mikelax:
@nazar  I added the migration and support for a JSON type for create and update character mutation. 
There is a bug when you try to edit a character, fields are not getting populated (label) and so the details form does not display. i think its related to the new renderprops added. 

I will work on adding support to validate the details json schema as part of mutation.

mikelax:
@nazar  I fixed a couple of errors related to the the edit character use case, though I am not 100% sure on the (render) props for the edit form. 

The label and profile image aren't loading (and possibly the details form) because I think [the _saveGameToState function was removed](https://github.com/mikelax/forgingadventures/pull/125/files#diff-d6e75b96c9c8d4f9070b69adf20e0d5eL89), and we didn't add back a way to get the values from the `character` prop into state. 

That function was called by `componentWillMount` and `componentWillReceiveProps`, I guess that is supposed to be handled by Formik, but I think we are missing the ability to do this little translation for these fields.

nazar:
Hey @mikelax ,

>There is a bug when you try to edit a character, fields are not getting populated (label) and so the details form does not display. i think its related to the new renderprops added.

We weren't exposing the character.labelId in the characters schema for the edit form to pick up. I just pushed a commit that fixes the edit form and improves the error messages so that it shows after a submit.

mikelax:
Thanks @nazar , going through changes now. 
two questions;  

1. Where does the `submitCount` variable come from? 
2. Do we also have to add code to handle the profile image url to get the image to diplay? That was also previously handled in the `_savegametostate` function?

nazar:
>@nazar see my previous comment, we technically already have this in the label.id

Yes, but it isn't in an easy to use shape for the form - so I just exposed it in the schema.

nazar:
>Where does the submitCount variable come from?

[Formik](https://github.com/jaredpalmer/formik/blob/4261887bbaea8a2313fee916e3f5e346e7cd8656/docs/api.md#submitcount-number)

mikelax:
>Yes, but it isn't in an easy to use shape for the form - so I just exposed it in the schema.

Gotcha, I was asking because I thought we were going to have to do special handling to get the profile image url set up, but I see you did that in the render function. so i guess we are good! 👍

nazar:
Hey @mikelax - I pushed a couple of visual tweaks (mostly to break up the form sections using Segments). Can you have a quick look and merge to master?

client/src/components/shared/components/RadioInput.js
```diff
@@ -0,0 +1,39 @@
+import _ from 'lodash';
+import React, { Component } from 'react';
+import { Radio } from 'semantic-ui-react';
+import PropTypes from 'prop-types';
+
+export default class extends Component {
+  static contextTypes = {
```
mikelax, line 7: @nazar  is this some crazy new react 16.3 magic??

mikelax (approved):
💯 🚀

client/src/components/shared/components/RadioInput.js
nazar: haha! That's the [old context API](https://5abc31d8be40f1556f06c4be--reactjs.netlify.com/docs/context.html) that Formik uses. I'll update this code to the new [16.3 React context api](https://reactjs.org/docs/context.html) once Formik updates to it as well.

Issue #127, opened 2018-05-02
Update game messages create component to add post type


mikelax:
Overview
===
As a user that already joined a game, I want the ability to create either an in character (IC) or out of character (OOC) post type when creating a new post in the game messages tab.  

Requirements
===
When creating a post, it should default to creating an in character post on initial page load, or after form resets. 
Above the text area, a “mock” post header should display to indicate to the user who the post is being created as.  It should display the profile image and name of either the user or character.  
There should be an “arrow” or UI element the user can tap to allow switching between user and character.  

A new attribute will have to be added to the posts model and db table to store the post type. 
Lets call the field:  postType not null
For the migration it should add the column as null, update existing rows in table to have value if ic, then as third step make column not null. 

Future enhancements 
===
In addition to allowing the user to change post type, we will also allow other actions to be made such as:  add dice roll, maybe image upload gets broken out into separate action, etc. 

Questions
===
Should postType be int or varchar or bool field?
I was thinking varchar with values of ic or ooc

nazar:
>Should postType be int or varchar or bool field?

enums ftw!

PR #128, opened 2018-05-04
Update game messages create component to add post type

nazar:
References #127 - WIP

- [x] add migration to support characters and posting types in game message
- [x] update server schema and client queries for new `game_messages` structure
- [x] update game message rendering to render characters and message types
- [x] update create game message to support choosing a character or OOC post type

### DIFF
--- a/client/src/components/Games/View/components/CreateMessage.js
+++ b/client/src/components/Games/View/components/CreateMessage.js
@@ -108,7 +108,7 @@ class PostAsSelectorBase extends Component {
           <Grid.Row centered>
             <Grid.Column>
               <Dimmer.Dimmable as={Segment} basic compact dimmed={this.state.ic}>
-                <Dimmer active={this.state.ic} inverted />
+                <Dimmer active={this.state.ic} inverted/>
                 <span>Out of Character</span>
               </Dimmer.Dimmable>
             </Grid.Column>
@@ -119,7 +119,7 @@ class PostAsSelectorBase extends Component {
 
             <Grid.Column>
               <Dimmer.Dimmable as={Segment} basic compact dimmed={!(this.state.ic)}>
-                <Dimmer active={!(this.state.ic)} inverted />
+                <Dimmer active={!(this.state.ic)} inverted/>
                 <CharacterProfile character={character}/>
               </Dimmer.Dimmable>
             </Grid.Column>
@@ -159,7 +159,7 @@ class PostAsSelectorBase extends Component {
   _getMyGamePlayerCharacter = () => {
     const myGamePlayer = this._getMyGamePlayer();
     return _.get(myGamePlayer, 'character');
-  }
+  };
 
 }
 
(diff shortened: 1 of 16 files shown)


mikelax:
Couple of issues found after testing:

1. Trying to update an OOC or IC post fails, the update ($input) is missing a value for `postType`.
Error:  Field value.postType of required type String! was not provided

2. When viewing the game messages as the GM, there are errors on the UI. 
One fix in `CreateGameMessage.js` function `CharacterProfile` is to use lodash to get the character.name. Even with this fix there are still other issues though. ie:
```js
function CharacterProfile(props) {
  const { character } = props;
  const characterUrl = _.get(character, 'profileImage.url');
  const characterName = _.get(character, 'name');
  const characterAvatar = characterUrl
    ? <Image avatar src={characterUrl}/>
    : <Icon name="user"/>;

  return (
    <React.Fragment>
      {characterAvatar} {characterName}
    </React.Fragment>
  );
}
```

client/src/components/Games/View/components/GameMessages.js
```diff
@@ -128,9 +140,53 @@ class GameMessageContainerBase extends Component {
...
+        <Comment.Content>
+          <Comment.Author>{gameMessage.user.name}</Comment.Author>
+          <Comment.Metadata>
+            <div>
+              Posted {this._relativeDate(gameMessage.created_at)}
+            </div>
+            {this._lastEdited()}
+          </Comment.Metadata>
+          <Comment.Text>
+            <RichEditor message={gameMessage.message} ref={c => (this.editor = c)} readOnly={!(editing)} />
```
mikelax: @nazar  should we refactor this `ref` to use the new react 16.3 `React.createRef()` 😮

client/src/components/Games/View/components/CreateMessage.js
```diff
@@ -42,22 +62,132 @@ class CreateMessage extends Component {
...
+    const myGamePlayer = this._getMyGamePlayer();
+    return _.get(myGamePlayer, 'character');
   };
+
+}
+
+function CharacterProfile(props) {
+  const { character } = props;
+  const characterUrl = _.get(character, 'profileImage.url');
+  const characterAvatar = characterUrl
```
mikelax, line 186: should you add `const characterName = _.get(character, 'name');` and use `characterName` below on line 188?

client/src/components/Games/View/components/GameMessages.js
nazar: 🤔 - looking up `React.createRef()`

client/src/components/Games/View/components/CreateMessage.js
nazar: Tempting but I think there will always be a `character.name` whereas `character,profileImage.url` is an optional field hence the `_.get` access.

mikelax (approved):
:shipit:

Issue #126, opened 2018-05-02
UI update on Game Messages posts display


mikelax:
Overview
===
 Now that we have a v1 of characters implemented we can revisit the game messages tab.  

As a user, when viewing the feed of game messages, I want to see posts that were made in character (IC) vs. out of character (OOC). The posts will be displayed in a single feed, but each post should render slightly differently to visually recognize the IC posts as more important. 

Requirements
===
IC posts can now have a proper header implemented. This means the avatar profile image can be displayed, and the real character name used. 

Add a concept so that the character header takes into account the game label. This will allow the UI to display additional game specific character fields in the header.  

For both pathfinder and dnd, let’s use this as a test case to also display in the header: 
Level, race, class
These can be displayed in parens in same line as name, perhaps though in a slightly smaller font?
Example:
Lindal (level 1 Halfling Bard)

For the OOC posts, we should render these to display as secondary to the IC posts. Perhaps the text is rendered in a light gray text?

Examples
---
https://rpggeek.com/article/28986731#28986731 - has "user made" banners, but these are created by each user, ie. not standard.
https://rpggeek.com/thread/1799806/tsname-ic-table-2-tales-yawning-portal/page/26
https://www.dndbeyond.com/forums/d-d-beyond-general/play-by-post/8583-the-forge-of-fury
- not that great as it looks more like a traditional forum, doesn't really have a "character" header

Future enhancements 
===
The user will have the ability to filter the thread to only display IC or OOC posts.

PR #131, opened 2018-05-15
#126 - UI update on Game Messages posts display - styling wip

nazar:


client/src/components/Games/View/components/CharacterLabelGameMessageHeaders/5e.js
```diff
@@ -0,0 +1,17 @@
+import _ from 'lodash';
```
mikelax: @nazar  cool!  one question, do you think that we should have a better naming convention to tie these headers back to the Game (labelId)? 
ie:  
1_dnd5e.js
2_pathfinder.js
etc

mikelax:
@nazar  when I try to view game messages for a 5e game, I am getting a null error, there are several ooc posts and one ic post currently in the game. 

```
TypeError: Cannot read property 'primaryLevel' of null
src/components/Games/View/components/CharacterLabelGameMessageHeaders/5e.js:11
```

mikelax:
@nazar  here are formulas to use for some of the header values.
Init:  Dexterity modifier. if `primary class` === `Bard` add 1/2 Proficiency (rounded down)

ie:  Level 1 Bard with Total Dex of 16 equals Init of +4
We would show the + sign for clarity

PR #136, opened 2018-05-27
132 Add gm post type to game messages table, create message.

### LINKED ISSUE #132: Game Master Posts are currently rendering as OOC posts, always
Overview
===
As the Game Master for a Game, when I am making a post I should be able to select the post type just like the Users can. My options should be:  OOC, or Game Master

If I make an `ooc` post, it should render just like the user ooc posts. 
If I make a `gm` post, it should render like a character post, ie. the "darker" UI display. 

Requirements
===
In the Rich Editor component, the GM should see a "slider" just like the users. Their options should be:  OOC, or "Game Master". 
If Game Master is selected, the value to user for the `postType` attribute in the messages table should be:  `gm`.

mikelax:
@nazar  I couldn't get the knex migration to add a value to the postType enum without dropping column, so I just modified the previous migration file. 
You can run the following two commands on your local db to make the change to the generated constraint:

```sql
alter table game_messages drop constraint "game_messages_postType_check";
alter table game_messages add constraint "game_messages_postType_check" CHECK ("postType" = ANY (ARRAY['gm'::text, 'ic'::text, 'ooc'::text]))
```

TODO
===
- [x] There is a perf issue introduced in this PR. when on the game messages tab, typing in the Message textarea is really laggy. Master is fine, so it is related to a change in this branch.

Closes #132

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -12,6 +12,21 @@ To lint this project run the following command:
 yarn lint
 ```
 
+## Database Migrations
+
+The project uses [knex for database migrations](http://knexjs.org/#Installation-migrations). Here are the basic commands to run:
+
+```bash
+# Create a new migration file
+yarn migrate:make issuenum_migration_name
+
+# From api container, run migrations
+yarn migrate:latest
+
+# From api container, Rollback last set of migrations
+yarn migrate:rollback
+```
+
 ## Configuration Notes
 
 The project uses the [node-config](https://github.com/lorenwest/node-config) package to manage configuration files between environments.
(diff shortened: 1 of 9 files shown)


nazar:
Hey @mikelax 

>I couldn't get the knex migration to add a value to the postType enum without dropping column, so I just modified the previous migration file

I'm sure I had this in a previous PR but I did find a way to add the enum migration. I'll see if I can update  the migration.

I'll also look into the perf issue 👍

nazar (approved):
🚢

Issue #123, opened 2018-04-11
Finish setting up api docker CFN client template


mikelax:
Overview
===
The react "client" CFN docker template is complete, working, and able to be deployed to test (and production. We need to finish the CFN template and associated resources for setting up the api docker app. 

Requirements
===

- [x] In api.yml, finish creating the Task Definition and Service
- [x] Create the SSM private key for test and production
- [x] Create production Dockerfile and webpack configs
- [x] Create the script to read/write encrypted keys from Parameter Store
- [x] Create docker script to read the keys on start up and make available to express via ENV variables
- [x] Create target group for API service, add listener rule for /api/*
- [x] Create new Wildcard SSL certificate and use that as default value for SSL ARN for test.yml file

mikelax:
Resolved in #137

PR #137, opened 2018-05-29
#123 - FA API Production Container

nazar:
Hey @mikelax - this PR is WIP but creating for discussion

References #123 

TODO:

- [x] Figure out Port and Security group for APITargetGroup - if 80 either need to alter existing client group and rename, or create new group and use that value for API CFN parameters, should also update README.


api/config/development.json
```diff
@@ -1,7 +1,14 @@
 {
   "auth0": {
     "faClient": {
-      "redirectUri": "http://localhost:3000"
+      "redirectUri": "http://localhost:3000",
+      "clientId": "i7Srpetd9dL9Dx1ckMIhBLFnSxksaVcz",
+      "domain": "forgingadventures-staging.auth0.com",
+      "audience": "https://api-test.forgingadventures.com"
+    },
```
nazar, line 8: Hey @mikelax 

I wasn't too sure if vars like `auth0. clientId` as secret so I tentatively moved them to the main `development.json` and `production.json` files and moved what I thought were secrets to `local-development.json` and `custom-environment-variables.json`.

Please advise if I need to rollback && refactor

api/config/production.json
```diff
@@ -0,0 +1,32 @@
...
+    "graphiql": false
+  },
+  "redis": {
+    "host": "redis"
+  },
+  "cloudinary": {
+    "apiKey": "514676315272514"
+  }
+}
+
```
nazar: @mike, please update the relevant IDs and Keys for the `production` env.

api/config/custom-environment-variables.json
```diff
@@ -0,0 +1,13 @@
...
+  "auth0": {
+    "managementClient": {
+      "clientSecret": "AUTH0_MANAGEMENT_CLIENT_SECRET"
+    }
+  },
+  "database": {
+    "connection": "DATABASE_URI"
+  },
+  "cloudinary": {
+    "apiKey": "CLOUDINARY_API_SECRET"
```
nazar: `apiKey` -> `apiSecret`

api/setEnvironmentSecrets.js
```diff
@@ -0,0 +1,50 @@
...
+  .catch((e) => {
+    console.log('Error', e);
+    process.exit(1);
+  });
+
+// private
+
+function getParams(nextToken) {
+  const params = {
+    Path: `/${process.env.PASSENGER_APP_ENV}/${process.env.APP_NAME}`,
```
nazar: switch app_name with env

api/etc/nginx/sites-enabled/nginx-forging.conf
```diff
@@ -0,0 +1,14 @@
+server {
+    listen 80;
+    server_name www.forgingadventures.com;
```
mikelax: @nazar I think this should be either `play` or `api`. 
The (prod) URL for the application will be play.forgingadventures.com
The question is should the web client make API calls back to `play.forgingadventures.com/api` or `api.forgingadventures.com`?

I think `api.forgingadventures.com` makes more sense, right?

nazar: Super thanks for the commit @mikelax 👍 

>I think api.forgingadventures.com makes more sense, right?

Yeah, I think for the express backend `api.forgingadventures.com` makes more sense. I have a feeling that we might need to setup CORS on express (which should be simplez) for the UI `play.forgingadventures.com` to make XHR calls to `api.forgingadventures.com`.

Oh, I just realised: do we need or plan on having a `test` or `staging` server?

mikelax: Yes we should have a test env. 
I will have to check the cfn scripts, but I know I have a domain for the client server (play) that is different for test vs prod
We should use the same format for api as well.

nazar:
Hey @mikelax - docker now builds the prod container. We just need to test running it 😅

I'm assuming that `APP_NAME` is set to `forgingadventures` and the env is `production`? I ran the secrets script but an empty results was returned. Are the params setup in the store? Maybe I'm using wrong path names?

api/Dockerfile
```diff
@@ -32,8 +32,8 @@ RUN rm /etc/nginx/sites-enabled/default
 COPY ./etc/nginx/main.d/env.conf /etc/nginx/main.d/secrets.conf
 COPY ./etc/nginx/sites-enabled/nginx-forging.conf /etc/nginx/sites-enabled/nginx-forging.conf
 
-# copy the environment setup script to /root - this will be used by /etc/my_init.d/30_secrets.sh
-COPY ./etc/my_init.d/30_secrets.sh /etc/my_init.d
+# copy the environment setup script to /root - this will be used by /etc/my_init.d/31_secrets.sh
```
mikelax: @nazar  I made this change, I assume the number just specifies the load order of the scripts, right?

infrastructure/api.yml
```diff
@@ -100,8 +100,10 @@ Resources:
       ContainerDefinitions:
         - Name: api
           Cpu: !Ref ContainerCpu
-          Command: [yarn start]
+          Command: [/sbin/my_init]
           Environment:
+            - Name: APP_NAME    # This var is used as namespace to get env. variables from Sys Manager Parameter Store.
```
mikelax: @nazar  does any other change need to be made to get this to carry through to the actual server? 

In the container logs,  I am seeing this error as the first entry in the log file. 

>APP_NAME or PASSENGER_APP_ENV variable is not set. Skipping secrets.

api/etc/my_init.d/31_secrets.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+if [ -z "$PASSENGER_APP_ENV" -o -z "$APP_NAME" ]; then
```
mikelax, line 3: @nazar  do we need to check for either of these envs, can we simplify to just looking for `$APP_NAME`?

infrastructure/api.yml
nazar: hey @mikelax - both `APP_NAME` and `PASSENGER_APP_ENV` need to be set as environment variables.

I think [this](https://github.com/WhoSay/dataslinger/blob/master/infrastructure/ng-client.yaml#L308-L312) is the relevant section from `matchplatform-ng-client` that we need to apply here.

infrastructure/apps.yml
```diff
@@ -118,6 +118,19 @@ Resources:
...
+      Actions:
+        - Type: forward
+          TargetGroupArn: !Ref ApiElbTargetGroup
+      Conditions:
+        - Field: path-pattern
+          Values:
+            - /graphql/*
+      ListenerArn: !Ref ElbHttpsListener
+      Priority: 30
+
```
nazar: Hey @mikelax - I'm going to have a little play and try to move both the `/graphql` and `/subscriptions` endpoints to be under `/api` for all envs (including dev). This might simplify our API infrastructure 🤞

api/Dockerfile
nazar: Nice catch @mikelax. Yep, the number prefix specifies the load order.

I changed 30 -> 31 as `passenger-nodejs` base container image already has something under `30`. I forgot to change the reference in `Dockerfile`

api/etc/nginx/sites-enabled/nginx-forging.conf
nazar: Hey @mikelax - I changed the above to `server_name localhost` as the API container is ENV agnostic. This also mirrors the Match ng/api nginx conf

api/etc/my_init.d/31_secrets.sh
nazar: 🤔 - I think checking for both makes sense otherwise the get secrets script will fail and prevent the container from starting... having said that, the container currently starts up but in development mode if none of the above ENVs are specified. Not sure if that's good or bad 🤔

infrastructure/apps.yml
nazar: This was [easier than I thought](https://github.com/mikelax/forgingadventures/pull/137/commits/c278bf1ecf5c4e6eae4356af544ce1c8942c88f1). Everything is now under `/api/*`.

I'll update `apps.yml` and remove the `/graphql/` rule

client/.env.production
```diff
@@ -3,5 +3,7 @@ REACT_APP_AUTH0_CLIENT_ID=AQ9U63ZdGNMfttXnztEiV5n7C8ZLwQSg
 REACT_APP_AUTH0_REDIRECT_URI=https://play.forgingadventures.com/login/callback
 REACT_APP_AUTH0_AUDIENCE=https://api.forgingadventures.com
 
+REACT_APP_WEBSOCKET_URI=wss://forgingadventures.com/api/subscriptions
```
mikelax: @nazar  I think the domain here should be:  `play.forgingadventures.com`

nazar:
I'm not sure if we need [these ARG and ENV entires](https://github.com/mikelax/forgingadventures/blob/master/client/Dockerfile#L20-L27). Maybe just an ARG for the `NODE_ENV`? Will test on my end.

nazar:
Hmm... re the client ENV variables. Looks like `.env.test` only applies when running tests via `yarn test` and does not apply if we want to deploy a test client to `playquest.forgingadventures.com` as CRA [always sets](https://github.com/mikelax/forgingadventures/blob/master/client/scripts/build.js#L4) the `NODE_ENV` to `production` during `yarn build`.

This might suggest that only the `.env.production` file is relevant. I'll dig into the `.dotnev` docs to see how we can define a `testing` or `playquest` env for deployment without having to specify all the ARGs during the Docker build stage.

nazar:
:shipit:

mikelax (approved):
🚀

PR #139, opened 2018-06-03
138 Fix setstate warning, turn on new eslint rule.

### LINKED ISSUE #138: Turn on eslint rule for "react/no-access-state-in-setstate"
Overview
===
This ticket covers the rule to add an eslint rule to ensure we are not making (obvious) error with the `setState` function. 

Requirements
===
The rule should be added to the `.eslintrc` file in the client folder. 
```json
"react/no-access-state-in-setstate": [2]
```

We then have to fix errors this rule picks up, currently there are 12 errors. 

Example code that needs to be refactored:
```js
  _formInput = (stateKey) => {
    return (e) => {
      const { store } = this.state;

      _.set(store, stateKey, e.target.value);
      this.setState({ ...this.state, store });
    };
  };
```

mikelax:
Closes #138 

- [X] Make fixes to `setState` calls. Notice most of the use cases required the same fix. I don't think there is a reason to use `...this.state`
- [X] fixed bug on Settings page where the data wasn't loading when transitioning from another page. It worked on fresh pageload. Fixed by updating to 16.3 lifecycle events.

### DIFF
--- a/client/src/components/Settings/index.js
+++ b/client/src/components/Settings/index.js
@@ -16,15 +16,6 @@ import { getProfile } from '../../services/webAuth';
 import { getMyDetails } from '../../actions/me';
 
 class Settings extends Component {
-  constructor() {
-    super();
-
-    getProfile()
-      .then(profile => {
-        this.setState({ profile });
-      });
-  }
-
   state = {
     activeItem: 'profile',
     displaySuccess: false,
@@ -38,6 +29,13 @@ class Settings extends Component {
     // the form validity state
     errors: {}
   };
+
+  componentDidMount() {
+    getProfile()
+      .then(profile => {
+        this.setState({ profile });
+      });
+  }
 
   static getDerivedStateFromProps(props) {
     return {
--- a/client/src/components/Settings/index.js
+++ b/client/src/components/Settings/index.js
@@ -5,7 +5,7 @@ import { graphql } from 'react-apollo';
 import { Helmet } from 'react-helmet';
 import { connect } from 'react-redux';
 import { compose } from 'recompose';
-import { Button, Container, Form, Grid, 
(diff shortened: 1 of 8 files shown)


client/src/components/Settings/index.js
```diff
@@ -16,7 +16,16 @@ import { getProfile } from '../../services/webAuth';
...
-  state = { 
+  constructor() {
+    super();
+
+    getProfile()
+      .then(profile => {
+        this.setState({ profile });
+      });
+  }
+
```
nazar: I'm wondering if the `getProfile()` block should go in `componentDidMount` as I think calling async functions in a constructor is discouraged.

Plus it saves is from calling `super()` which can have gotchas if `this.props` ever need to be accessed in the constructor.

client/src/components/Settings/index.js
```diff
@@ -30,22 +39,15 @@ class Settings extends Component {
...
-  componentWillMount() {
-    return getProfile()
-      .then(profile => {
-        this.setState({ profile });
-      });
-  }
-
-  componentWillReceiveProps(nextProps) {
-    this.setState({
+  static getDerivedStateFromProps(props) {
```
nazar: NICE! 👍

client/src/components/Settings/index.js
mikelax: @nazar  ah, well i did see that i should be passing `props` in the super. 
but i will give it a shot. 

In [the docs](https://reactjs.org/docs/react-component.html#constructor), are you referring to this sentence with regards to having async functions in constructor?
>Avoid introducing any side-effects or subscriptions in the constructor. For those use cases, use componentDidMount() instead.

nazar (approved):
🚀 🚢

PR #141, opened 2018-06-04
#135 - Add sub routes to Games view

### LINKED ISSUE #135: Refactor Games Page
1. Move the Game Lounge tab to it's own endpoint - `/games/:id/messages`
2. Move Game Messages tab to it's own endpoint -  `/games/:id/lounge`
3. Move the Players tab under the game overview in `/games/:id`
4. Remove tab controls from '/games/:id`

nazar:
Closes #135


mikelax:
hey @nazar  this is looking *MUCH* nicer. just a couple of tiny things that I found. 

- When I am viewing the game details as a player or GM. the "settings" icon is really small in the Game Actions column. It's tiny compared to all other elements in the table. It should be made a bit larger.
  - On a similar note the size of the icon and text in the Status column is really larger, that could be made a bit smaller. It doesn't need to be any larger than the Character name text. 
- After I have visited all three tabs (details, lounge, messages), switching to either the lounge or messages pages takes over a second to load. I can see in the console though no additional queries are being sent to server. When I switch to the details page that switches right away.

nazar:
>switching to either the lounge or messages pages takes over a second to load

Yep - I noticed that as well. I'll 👀  into it 👍 as I can't think of any reason why it shouldn't* render instantly with a loading indicator whilst the graphQL query is in flight.

mikelax:
@nazar  at first i thought it was slowness related to db queries, graphql, etc but its all loaded in apollo cache already so it must be something related to react lifecycle. i will also try to take a peak if i get a minute.

nazar:
Thanks for taking a look @mikelax - I've narrowed it down to the `cloudinary-react` component. If I comment out the in character message header then the page loads and switches instantly when clicking into the Game Messages view. :digging: deeper.

nazar:
[This](https://github.com/cloudinary/cloudinary-react/issues/26) is what we're seeing :sadpanda: - I'll see if using the cloudinary URI directly improves the rendering

nazar:
@mikelax - fixed both issues

mikelax (approved):
:shipit:

PR #142, opened 2018-06-05
closes #130 - Make typing after Quote easier

### LINKED ISSUE #130: UX improvement - Make typing after Quote easier
Overview
===
There is a UX issue right now if the user tries to quote a message. 
After clicking the quote link, the message is inserted into the RichEditor with the blockquote styling. This is all correct. 
The hard UX is if the user wants to type more text after the quoted block, which will certainly be the common use case. Currently, the user must hit return to create a newline, then click the blockquote style button which will remove the styling from that newly created line. 

Instead of this, when a user hits the quote link, after inserting the message into the RichEditor, we should automatically create the newline and remove the blockquote styling from that new line.

nazar:
Hey @mikelax - some notes for this PR.

* I added TinyMCE scripts and plugins to the public folder as we were grabbing these off tiny's CDN but we were not able to pin the tinymce version. My concern is that we didn't have control of tiny's update cycle, which I feared might break our integration in one of their future updates. For reference, the `stable` version of tinymce's JS we were using was last updated about two weeks ago. What do you think about this? Should we include tinymce's JS and plugins locally or continue to rely on their CDN?

* I made both the lounge and message editors sticky (via CSS) so that the editor is always visible. My only concern is with the Game Message editor which is a bit tall and obscures half the screen on a 1080p display. What do think? Should we keep the sticky behaviour and maybe reduce the Game Message editor height or just drop the sticky?

* I fixed an I was seeing an error in the Almost finished  view when type in the `name` field

Closes #130


mikelax:
Great, thanks @nazar. 
Let me think about local tiny scripts 
WRT sticky editor, I don’t think that it should be sticky. I think it’s too big like you said, even on mobile. 
I was actually wondering if we should move the editor so its at the bottom, ie, after the last message (at least for the messages tab). This is how you would typically do things. You would read through the posts, then you write a new message. Kind of like slack, or even GitHub here as the editor is at the bottom.

nazar:
>wondering if we should move the editor so its at the bottom

👍 👍 - done

nazar:
This needs a few more tweaks

nazar:
Looks better now

mikelax (approved):

PR #143, opened 2018-06-07
Implement dice roller for RichEditor and Game Message

### LINKED ISSUE #75: Implement dice roller for RichEditor and Game Message
Overview
===
One of the required features in order to allow users to play games is a dice roller. This will allow the user to create one or more rolls using commonly accepted dice syntax. 

Examples
===
[Here is the BGG dice roller](https://boardgamegeek.com/wiki/page/Geek_Randomizer_Guide) guide. You can see it supports a lot of different formats. Some of the commons one are: 

* 2d6+3 - roll 2 6 side and add 3
* 2d20l+3 - roll 2 twenty side and drop the lowest and then add 3
* 1custom3{Tom;Dick;Harry} - roll 1 custom 3 side, using the values Tom, Dick, Harry
* Exploding dice
* Fudge dice

Existing Packages
===
* [chance](https://www.npmjs.com/package/chance) - This is a mature package that focuses in trying to get close to real random. The `rpg` and `dice` rolls are only a really small part of the overall package. Probably not focused enough on just dice rolls. 
* https://www.npmjs.com/package/d20
* https://www.npmjs.com/package/roll

Out of the above libs, I think the [roll](https://www.npmjs.com/package/roll) package is closest to what we are looking for. 

Requirements
===
* A User should be able to add multiple dice rolls to a single message. 
* The User should be able to add a descriptive label to each roll.
* Once a message is saved, the dice rolls should be able to be edited (ie. they are immutable).
  * A nice to have would be to not allow the user to remove any dice rolls once saved, though I don't know if this is required for MVP. 
* Add a new column to game_messages table to store an array of dice rolls. Each dice roll object will be an object of:  label, result, components of roll
* We do only need to implement some of the basic dice modifiers as this feature certainly follows the 80/20 rule. 20% of the dice features will be used 80% of the time. 
These features are:  
* multiple dice, add/subtract constants
* drop high / drop low
* custom dice

Future Considerations
===
Once we have some basic character attributes we can use those values to automatically calculate certain rolls. These rolls will have to be based on the game label. For example for d&d5e, to roll `Initiative` is 1d20 plus the character `Dexterity` modifier. So the user could say:  `/roll Initiative`

We will only be able to easily calculate certain rolls. We could also allow the user to create their own roll templates, or bookmarks. For example:  `/roll 2d20l+3 save advantage-attack`
they could then later use:  `/roll advantage-attack`

nazar:
Hey @mikelax 

Creating the PR super early to follow progress and for discussion.

What do you think of replacing Joi on the API with Yup? That might make it easier for us to define validation on both sub-projects without having to switch contexts between the two formats. Tbh, I prefer Yup as it is simpler and Joi does have some weirdness with booleans and dates.

Yup supports custom types so we should be able to use the `dice` library to validate inside Yup on both the client and api.

I was also thinking of linking the `game_messages.meta.rolls` array Yup validator to the label character (at least in the same file) - this should give us flexibility to extend each label's dice roll validations.

Closes #75


mikelax:
hey @nazar , this is great! 
👍 for switching from joi to yup on the api. not sure why we didn't do that when we implemented characters.

mikelax:
hey @nazar  I just did some testing and this is really close!!!!! It looks great, here are a couple of tiny things I found that we may want to fix. 

I also pushed a few small updates including some css, feel free to modify, by changing heirarchy, etc. 
Also for some reason my client `yarn.lock` was updated and that got committed as well. 

1. In the dice overlay, when you click on a dice button (ie d20) it clears any text already in the textfield before adding "d20". Instead the functionality should be to add the new die where the cursor is in the textfield, or if that isn't possible then append to the end of the current value. 
1. Open the dice overlay and add one or more dice rolls. Submit the overlay. Open the dice overlay again and create another one or more rolls. After submitting the overlay the second time, notice the original list of rolls are overwritten by the second set. This list should be additive. So the new rolls keep being added to the bottom of the list. 



There is a js error when viewing Game Messages, I have created a few messages that contain rolls. 
```
bundle.js:168002 Warning: React does not recognize the `computedMatch` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `computedmatch` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    in div (at index.js:31)
    in Switch (at index.js:27)
    in div (created by Container)
```

nazar:
Hey @mikelax 

>Instead the functionality should be to add the new die where the cursor is in the textfield

👍 

>This list should be additive. So the new rolls keep being added to the bottom of the list.

👍 

>React does not recognize the `computedMatch` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `computedmatch` instead

I think that is due to [this](https://github.com/mikelax/forgingadventures/blob/master/client/src/components/Games/index.js#L31) div being nested inside the 'Switch' component. I'm not sure how to refactor this so I'm ignoring it for now as it is only a warning.

mikelax:
@nazar  I think this is ready to :shipit: 
I just made a couple of edits. I allowed the dice rolls to be done by anyone, not just `ic` post types. As we needed `gm` at a min, but there is no reason to not allow `ooc` also roll. 

Also brought the private message render functions closer together. I think those can be refactored to be a single function as they are 95% the same, only difference is the header and the className.

mikelax (approved):

nazar:
Hey @mikelax 

I ❤️ your styling of the roll results!! Let's :shipit:

mikelax:
thanks @nazar merging now. 

Also, I just updated the description of this PR, but we can use the [following words to auto-close](https://help.github.com/articles/closing-issues-using-keywords/) the associated issue with the PR.

nazar:
>but we can use the following words to auto-close the associated issue with the PR.
👍  - I used that as a commit message in one of the previous commits, which also auto closes the ticket on merge.

PR #146, opened 2018-06-12
closes #108 - Implement Custom Styles for Game Message Formats

### LINKED ISSUE #108: Implement Custom Styles for Game Message Formats
Overview
===
As a Player in a Game, I want to be able to easily select from pre-formatted styles for common formats such as:  Talking, Thinking, Singing.

Requirements
===
These custom styles will be implemented with tinymce [custom formats](https://www.tinymce.com/docs/demo/format-custom/)

Here is the list of formats that we should have. This should be listed in the menu Dropdown in this order. 

* Paragraph - this is the regular, default text that will be used in textarea
* Header - This is a message header, it can be formatted as the current Header 2
* Character Talking - This will be the character talking text. It should be bold, and have a color to it.
* Character Thinking - This is character thinking, it should be colored, bold, and italics. 
* Character Singing - This is singing text, it should be colored, italics, and indented one group.

**NOTE**, in a future version, we can allow the user to select a "Color Palette" for their messages, and this will then be used to format (color) IC message formats such as Talking, Thinking, and Singing.

The existing `Header 3` format can be removed, and `Header 2` can be renamed to `Header`

nazar:
Closes #108


mikelax:
hey @nazar  this is looking good. Here are a couple of quick callouts that I noticed. 

* When editing a game message, the Formats dropdown is missing.
* Can we change what displays in the formats dropdown from "Formats" to "In Character Formats"
* When a format is selected, can that given format be displayed in the dropdown. Its a little confusing now as for example the "B" icon is displayed as on, but you have to click again the Formats dropdown to see that "Talking" is selected.
* I think that we should not display the icons (talk, think, sing) in the editor. It's too confusing with the cursor position vs. the position of the icon. I think it's cool to display them in the thread though. One change would be that the talk and think icons should display both before and after, just like the sing icons so they are consistent. 

**This next one I am on the fence about.**
If you have a Format selected (ie. Talking), and then hit the enter key. should that "end" the Format? ie. this is how the "Header" works. Don't make any change for this yet but give it some thought and we can discuss.

nazar:
>When editing a game message, the Formats dropdown is missing

fixed

>Can we change what displays in the formats dropdown from "Formats" to "In Character Formats"

>When a format is selected, can that given format be displayed in the dropdown. Its a little confusing now as for example the "B" icon is displayed as on, but you have to click again the Formats dropdown to see that "Talking" is selected

I looked through the tinyMCE docs for the `styleselect` toolbar control but couldn't find anything relevant for the above. Reading one SO answer suggests we might need to implement our own custom plugin to support this.

>If you have a Format selected (ie. Talking), and then hit the enter key. should that "end" the Format?

I did add an option to apply the above in anything selected in the "paragraph" and "heading" list. This same reset, however, does not seem to be applying to anything selected from the Formats list.

We might need to dive deeper into the docs as the "Formats" menu might not be the best place to put our custom formatting? We just need to figure out how to add custom items to the `formatselect` toolbar list control.

nazar:
I might be able to copy [this](https://github.com/tinymce/tinymce/blob/48a9d567c5ae2c215d812301d36f7c66adf0fd08/src/ui/main/ts/editorui/FormatSelect.ts) `formatselect` plugin, which is used to show the "header" and "paragraph" options and extend it to add our own custom formats.

Maybe something to try in v1.x?

nazar:
>When a format is selected, can that given format be displayed in the dropdown. Its a little confusing now as for example the "B" icon is displayed as on

I might have a solution for this....

mikelax:
@nazar  cool, with the latest updates I think that we are in good shape. If there is an easy way to "not sync" the B, I, buttons when a Format is selected I think that will be good to call it for v1. 

In a point release we can look to see about a custom plugin, etc. If we do switch up the Formats, that won't affect the styling for any already created posts, right?

nazar:
> If there is an easy way to "not sync" the B, I, buttons when a Format is selected I think that will be good to call it for v1

My last push fixed that 😉

nazar:
>In a point release we can look to see about a custom plugin, etc. If we do switch up the Formats, that won't affect the styling for any already created posts, right?

Yeah, I'm concerned about that too. I updated the format styling to only set class names so that 🤞 should minimise plugin transition issues...

mikelax (approved):

PR #151, opened 2018-06-19
closes #149 - Refactor Game Details page to use existing Profile Image Component

### LINKED ISSUE #149: Refactor Game Details page to use existing Profile Image Component
Overview
===
On the Game Details page (/games/:id) we are showing a list of the users and characters for the given game. The user profile images and character profiles images should be refactored to use the existing profile image component that is used elsewhere, ie. game lounge and game messages. 

We have to handle the one use case for the "still working on it" for the character. It would be nice if the icon (currently an exclamation point) is the same size as the profile images so things look consistent. 

Oh yeah, requires blowing away the existing DB via `docker rm forgingadventures_db_1`

nazar:
Hey @mikelax 

It was proving tricky trying to size fonts (and icons) to have the same behaviour as % size images so I uploaded a couple of placeholder profile/character images to Cloudinary (I tried to go for something androgynous but feel free to replace and update the image's publicId in the [.env](https://github.com/mikelax/forgingadventures/blob/149_profiles/client/.env) file.


client/src/components/shared/ProfileImageAvatar/index.js
```diff
@@ -35,32 +35,39 @@ function CloudinaryImageAvatar(props) {
   const { publicId, size } = props;
   const cloudName = process.env.REACT_APP_CLOUDINARY_CLOUDNAME;
   const width = {
-    normal: 80
-  }[size];
+    small: 80,
+    tiny: 45
+  }[size || 'tiny'];
```
mikelax: @nazar  I don't think this snippet `size || 'tiny'` is working. 
Currently when i was logged in with a user that had a cloudinary url, i saw a broken image in the header, it was because the header was sending in a `size` value of `mini`. I think width should be (not sure about the 45 value for mini):  

```
const width = {
    small: 80,
    tiny: 45,
    mini: 45
  }[size];
```

client/src/components/shared/ProfileImageAvatar/index.js
```diff
@@ -35,32 +35,39 @@ function CloudinaryImageAvatar(props) {
...
+
+  const url = `https://res.cloudinary.com/${cloudName}/image/upload/c_crop,g_auto,h_${width*4},r_max,w_${width*4}/c_fill,w_${width}/${publicId}`;
```
mikelax: @nazar  there is something wrong with the width value used here vs. what is being displayed on the page. 
On the game details page, the logic works, a 45px image is displayed at 45px. 

on the lounge and messages pages though, the image is displayed at 80px but the source image is only 45px. 

Also, I am not sure if the logic for a `c_fill` after the `c_crop` is correct. For example look at this image, the focus is just my eye. That started happening with the time before this was refactored. 

screen shot 2018-06-19 at 4 44 10 pm

What about something like this instead? replace `c_crop` with `c_fill`:
https://res.cloudinary.com/forgingadventures/image/upload/c_fill,g_auto,h_180,r_max,w_180/w_45/SkxvtMSDz

client/src/components/shared/ProfileImageAvatar/index.js
```diff
@@ -35,32 +35,39 @@ function CloudinaryImageAvatar(props) {
...
   const { size } = props;
-  const fontSize = {
-    normal: '40px'
-  }[size];
+  const publicId = process.env.REACT_APP_CLOUDINARY_AVATAR_PLACEHOLDER;
+
+  return (
+    <CloudinaryImageAvatar publicId={publicId} size={size} />
+  );
+}
```
mikelax, line 63: @nazar  for both this function and `NoProfileIcon` functions, what do you think about adding prop types and a default prop for the `publicId`? 

You can do that even with [stateless functions](https://github.com/yannickcr/eslint-plugin-react/issues/237)
something like: 

```
NoAvatarIcon.propTypes = {
  size: PropTypes.string.isRequired,
  publicId: PropTypes.string
};
NoAvatarIcon.defaultProps = {
  publicId: process.env.REACT_APP_CLOUDINARY_AVATAR_PLACEHOLDER
};
```

nazar:
```javascript
NoAvatarIcon.propTypes = {
  size: PropTypes.string.isRequired,
  publicId: PropTypes.string
};
```

I should be adding `propTypes` more. I'd add these to public components (i.e. ones the are exported).

```javascript
NoAvatarIcon.defaultProps = {
  publicId: process.env.REACT_APP_CLOUDINARY_AVATAR_PLACEHOLDER
};
```

I'm not too keen on the `defaultProps` usage here as the constant is defined outside the body of the function, which itself is small. In this instance, I prefer it the way it is.

client/src/components/shared/ProfileImageAvatar/index.js
nazar: Oh wow! Is `c_crop` zooming into the eye?? `c_fill` looks 10000% better!

nazar: >on the lounge and messages pages though, the image is displayed at 80px but the source image is only 45px.

Nice catch @mikelax 👁 - looks like the cloudinary example I copypastaed first did a crop then a fill. I've tweaked it a little now just to do a fill

mikelax (approved):
🚀

PR #152, opened 2018-06-19
closes #140 - warning in controlled element on Settings page - email field

### LINKED ISSUE #140: Error in controlled element on Settings page - email field
Overview
===
On the /settings page, when you load the page (either fresh pageload or page transition) there is a js error on the page. 

The error relates to switching from an uncontrolled to controlled element, related to the email field. I *believe* it's because there isn't an `onChange` attribute on the emai field since it's read-only?

nazar:
This was a warning and is triggered since the email input is initially rendered with a null email value (un-controlled) and again once an email is fetched to the db (controlled).

I updated the code so that the email input is only rendered when there is an email

### DIFF
--- a/client/src/components/Settings/index.js
+++ b/client/src/components/Settings/index.js
@@ -109,18 +109,20 @@ class Settings extends Component {
                     </Message>
                   </Form.Field>
 
-                  <Form.Group inline>
-                    <Form.Input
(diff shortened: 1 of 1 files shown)


mikelax (approved):
:shipit:

PR #154, opened 2018-06-19
Snake case database columns with a knex snake -> camel case transformer

nazar:
Hey @mikelax 

I've been thinking about and toying with this for a while. The main issue this PR addresses is being able to define table and column names using snake_case but using camelCase identifiers in the UI. My motivation is that both PG and mySQL require special handing of mixed case column names (with the later requiring a switch at the db config level)

Both knexjs and objection support column name transforms and objection [provides a snake <-> camel case mapper](https://vincit.github.io/objection.js/#objection-knexsnakecasemappers). This PR only affects the migrations. The rest of the API and the UI remains un-affected.

Remaining Tasks
---

- [x] See if it's easy to create database and ensure all objects are created there instead of `public`


mikelax:
hey @nazar  , I did a bunch of reading here and it looks like the postgres docker container does (by default) create a `postgres` database, as viewed in [the docs here](https://docs.docker.com/samples/library/postgres/#postgres_user). 

So long story short, I think it's a no-op for getting db support. we would just have to update the db name in the db conn string (stored in parameter store) for test/production environment when we are ready. so i will give this PR a try shortly (after dumping the data).

nazar:
Super thanks for looking into this @mikelax - I did have a look through the knex docs and although we can specify a `database`, we cannot use knex to create the database itself.

It's one of those chicken/eggs thing. Knex attempts to create it's metadata tables on first run, which happens before any migrations are run - even if the first migration creates a database.... 

I'll have one more look incase I missed anything.

mikelax:
@nazar that’s right. But I think it’s fine for what we are doing. For dev we have the Postgres db created automatically. 

For test and prod, the one time we would have to create the db manually after the infrastructure CFN stack is spun up but before the apps are deployed.

mikelax (approved):
💥

PR #156, opened 2018-06-20
Other small updates from camel -> snake case

mikelax:
After more testing, there were a few more changes we had to make to finish this PR. 

- We had to fix all the `created_at` and `updated_at` fields since they now were being mapped to camelCase. 
- For searching games, I had to update the query that was using json search in the `gameSettings` field to explicitly be `game_settings`
- Unrelated, but I fixed a bug in the gamePlayer mutation `updateGamePlayer`


api/src/schemas/gamePlayer.js
```diff
@@ -100,9 +100,7 @@ export const gamePlayerResolvers = {
     updateGamePlayer: (obj, { id, input }, context) =>
       schemaScopeGate(['create:posts'], context, () => {
         return getOrCreateUserByAuth0Id(context.req.user.sub)
-          .then((user) => {
-            input.userId = user.id;
-
+          .then(() => {
```
mikelax, line 103: @nazar  I removed setting the `userId` on the input for the update. The userId should already be there, and this was causing a bug where if the GM "accepts" a player, the userId value of the player was getting updated to the GM.

api/src/services/games/getGames.js
```diff
@@ -44,13 +44,13 @@ function doGameSettingsSearch() {
     const { skillLevel, postingFrequency, gameStatus } = gameSettings;
 
     if (skillLevel) {
-      this.gameQuery.whereJsonSupersetOf('gameSettings', { skillLevel });
+      this.gameQuery.whereJsonSupersetOf('game_settings', { skillLevel });
```
mikelax, line 47: @nazar  I don't know why, but the games search didn't work until I change the column to use the snake case. I think it's because `whereJsonSupersetOf` is changing the statement to a database cast function.

nazar: Oh super thanks for these @mikelax 🚀

nazar (approved):
🚀 🚀 🚀

PR #157, opened 2018-06-20
Misc be sure to also copy over .env file.

mikelax:
Simple change, one question @nazar  how come in the similar DS change the command was: 
`COPY .env.* /usr/src/app/` notice the different destination as opposed to `.` here?

*edit*:  is it because of this command a few lines up? `WORKDIR /usr/src/app`


nazar:
>is it because of this command a few lines up? WORKDIR /usr/src/app

👍

nazar (approved):
🚀

Issue #153, opened 2018-06-19
Implement Audit Trail History for Character Changes.


nazar:
entries should include:  timestamp, user making change, field deltas, optional Game ID, maybe even Game Message.

nazar:
Hey @mikelax - Some thoughts on the character audit trail feature plus how it will be used when creating and displaying game messages and character headers.

**Table structures**

![tables](https://user-images.githubusercontent.com/6967/41714202-70b5f01a-7547-11e8-8b3d-2767c6ae0e4a.png)

**character_logs**

Records any change to `characters.character_details`, starting from character creation. Additionally, a `change_description` column is added to record reasons (i.e. character born, HP loss from damage, etc...)

**characters**

We add a `last_character_log_id` column to hold the last recorded change. This is slightly denormalises the table relationship but it makes it quicker and easier to lookup the last `character_logs` entry for a given `character_id`

**game_messages**

This table is modified to record the `character_log_id` instead of the `character_id`. Each time a `game_messages` record is created, we use the `character_log_id` from `characters.last_character_log_id`

**A pretty sequence diagram to illustrate logic**

![fa character logs](https://user-images.githubusercontent.com/6967/41714632-e2db8f14-7548-11e8-8056-457bae659611.png)

mikelax:
hey @nazar  great work. I think this is just about there, and totally makes sense. Here are a couple of small comments. 

* I would still have a `characterId` column in the game_messages table. I know it de normalizes, but I think it could aid in looking up character details, ie. profile image. I assume character profile image will always shows the most recent as that will be a quick visual indicator when scrolling through messages who posted what. 
* Add a fk to the character_logs table that will link to the prev log id. The initially created log entry will have a null entry for this field.

PR #158, opened 2018-06-22
Refs #153 - Implement Audit Trail History for Character Changes

nazar:
Opening this for preview/testing - API level validations for character create/edit are still wip


mikelax:
hey @nazar  man, you are getting fast on this work!!! 
I was thinking about this task this am while walking the dog. Here are a couple of thoughts. 

- Should we break up the existing `character_details` column into two or more fields? This would allow us to more easily add audit history for only a set of fields. ie. if the user updates the `Background Information`, or adds an Item (still need to add Equipment field) should this cause a snapshot? 

After typing and thinking about it a bit more, maybe that is actually fine, it potentially will eat up a bit of storage, ie. if you update a single number field (xp) that will snapshot all the textareas like background, etc. What do you think? 

Do we need to worry about "versions". ie. let's say we roll out this PR, then next week we add two more fields to the character, what happens now that there could potentially be rows in the history table that don't match. Or would adding a new field in the migration update any records in the history table as well?

nazar:
>After typing and thinking about it a bit more, maybe that is actually fine, it potentially will eat up a bit of storage, ie. if you update a single number field (xp) that will snapshot all the textareas like background, etc. What do you think?

Good point! I thought about storing diffs or partials in the table but was slightly concerned that might limit our options in how to display differences in the future. Storing a snapshot is the simplest for now and as for DB storage, I think PG has a set size per db table so the gain, if any, between storing a diff vs snapshot are negligible. 

>Do we need to worry about "versions". ie. let's say we roll out this PR, then next week we add two more fields to the character, what happens now that there could potentially be rows in the history table that don't match. Or would adding a new field in the migration update any records in the history table as well?

I think we have to be careful how we handle versions as there will be issues, like you pointed out, if we update a label's character fields - ideally this shouldn't happen mid game but might be unavoidable.

nazar:
Hey @mikelax - this PR is ready for testing and merging. Please note that a database wipe will probably be required due to the new FK relationships.

I've made some assumptions that data will be available - i.e. `characters.character_details.meta` and `character_logs.character_details.meta` so the migrations cover only schema and not data changes.

nazar:
closes #158 and #153

mikelax:
@nazar  just did some testing, found the following issues:  

1. As the GM, I can not create a game message. I receive this error if I try to post as either the GM, or OOC GM. 
Got the following error:
```
Uncaught (in promise) Error: GraphQL error: Cannot destructure property `labelId` of 'undefined' or 'null'.
    at new ApolloError (:3000/static/js/bundle.js:6300)
    at Object.next (:3000/static/js/bundle.js:4971)
```

nazar:
Nice catch @mikelax - :hawkeye:

mikelax (approved):
👍 
boom, nice work. this looks great!!

PR #159, opened 2018-06-26
add summary queries and breakup profiles page into routes

nazar:


api/src/schemas/character.js
```diff
@@ -80,6 +86,16 @@ export const characterResolvers = {
...
           .orderBy('updatedAt', 'desc');
       });
+    },
+    myCharactersSummary: (obj, variables, context) => {
+      return runIfContextHasUser(context, (user) => {
+        return Character
+          .query()
+          .select({ id: 1 })
+          .count({ charactersCount: 1 })
+          .where({ userId: user.id })
```
mikelax, line 96: @nazar  do we need to add where clause for `deletedat` is null or does knex handle that automatically?
same question goes for all the other summary counts.

mikelax (commented):
one tiny comment, otherwise this can be merrged. 🚀

api/src/schemas/character.js
nazar: Good question @mikelax - there is an [objection plugin](https://github.com/griffinpp/objection-soft-delete) that handles that - we can mix that into any model we want to enable soft deletes on.

PR #161, opened 2018-06-28
Add and Improve Paging

### LINKED ISSUE #160: Add and Improve Paging
Revisit paging in the games page and use paging as described [here](https://blog.logrocket.com/rendering-large-lists-with-react-virtualized-82741907a6b3). Ideally, we should use cursor based paging - revisit graphQL docs for any pointers.

Implement paging in:

* game lounge
* game messages

nazar:
Closes #160


mikelax:
hey @nazar , found a couple of small issues with this. 

1. When viewing the game messages page with 0 messages created, there is an error. 
JS error on page: 
```
index.js:2177 Warning: Failed prop type: UberPaginator: prop type `perPage` is invalid; it must be a function, usually from the `prop-types` package, but received `undefined`.
    in UberPaginator (created by Route)
    in Route (created by withRouter(UberPaginator))
    in withRouter(UberPaginator) (at index.js:46)
    in GameMessages (at ViewGameMessages.js:23)
```

The graphql error in the console. 
```
db_1      | ERROR:  OFFSET must not be negative
db_1      | STATEMENT:  select "game_messages".* from "game_messages" where "game_id" = $1 order by "created_at" asc limit $2 offset $3
api_1     | select "game_messages".* from "game_messages" where "game_id" = $1 order by "created_at" asc limit $2 offset $3 - OFFSET must not be negative
api_1     |
api_1     | GraphQL request (2:3)
api_1     | 1: query gameMessages($gameId: ID!, $page: Int, $perPage: Int) {
api_1     | 2:   gameMessages(gameId: $gameId, page: $page, perPage: $perPage) {
api_1     |      ^
api_1     | 3:     ...gameMessagesFields
```


Once there were two messages in the list I still received the same error message, but the parameter value for `page` was 0 instead of -1. 
```
index.js:2177 Warning: Failed prop type: UberPaginator: prop type `perPage` is invalid; it must be a function, usually from the `prop-types` package, but received `undefined`.
    in UberPaginator (created by Route)
    in Route (created by withRouter(UberPaginator))
    in withRouter(UberPaginator) (at index.js:46)
    in GameMessages (at ViewGameMessages.js:23)
```

2. After the above error, I then created the first message, it happened to be from the GM. This caused another error, ie. the new message was not added to the list. After I refreshed the page though the message was displayed. I then tried to add a second message and that was added to the list correctly, so there is an edge case around creating the first message for a game (probably the same for Lounge as well).

mikelax:
another note, finally got another messages posted to create a second page of messages. 
One small issue, I don't think the UI is handling the use case where a message crosses the threshold to the next page. ie. posting the 26th message. 
I had to do a page refresh to get the paging controls to display and show the second page of data (previously everything was displaying on page 1 with no paging controls).

nazar:
>One small issue, I don't think the UI is handling the use case where a message crosses the threshold to the next page. ie. posting the 26th message.

I should have mentioned that the pagination only renders on a page load. New messages that cross the page threshold still show on the same page - I didn't add support for dynamic client side pagination as it wasn't trivial.

mikelax:
>I should have mentioned that the pagination only renders on a page load. New messages that cross the page threshold still show on the same page - I didn't add support for dynamic client side pagination as it wasn't trivial.

OK, that makes sense for that issue. I think its fine.

nazar:
🤦‍♂️

mikelax (commented):
@nazar  :shipit:

mikelax (approved):

PR #162, opened 2018-06-28
Migrate to universal auth0 login page.

mikelax:
If this looks good I will do a final commit to remove the commented out code.


client/src/services/login.js
```diff
@@ -1,47 +1,66 @@
-import Auth0Lock from 'auth0-lock';
+// import Auth0Lock from 'auth0-lock';
+import Auth0 from 'auth0-js';
```
mikelax: @nazar  all the examples use `import auth0` , should I change this back from capital `Auth0`?

PR #164, opened 2018-06-29
closes #155 - set a max height on the game description

### LINKED ISSUE #155: Need to limit display of game overview field on games list page
Overview
===
If a game contains a really long game overview, the entire overview is displayed in the game card on the list games page. We need to use some elipses to limit the display of this field so the cards all have a standard and uniform height. 

here is an overview to show the issue.
```
The road on which you travel soon branches into a four- way crossroads, forming an elongated x-shape. Various shops and stores line these streets, but your attention is drawn to a large, well-kept inn on one of the triangular parcels of land formed by these roads. A freshly painted sign in the shape of a plump fish bears the name of this interesting establishment: The Whistling Fish, Potentate of Potations.



The dungeon will be down the road, a day's walk outside of town.
```

nazar:
Hey @mikelax 

It's tricky getting ellipses to work on multiline text, more so if the text has nested components like we do in the `description` field, which is produced by tinyMCE and contains nested `p` elements for paragraphs.

I've limited the description to four lines for now.

### DIFF
--- a/client/src/components/Games/components/GameCard/gameCard.styl
+++ b/client/src/components/Games/components/GameCard/gameCard.styl
@@ -1,0 +1,12 @@
+.game-card-component
+  .truncated
+    max-height 80px
+
+    overflow: hidden;
+    text-overflow: ellipsis;
+    -webkit-box-orient: horizontal;
+    display: -webkit-box;
+    -webkit-line-clamp: 4;
+
+    & > *
+      line-height 20px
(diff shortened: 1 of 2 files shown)


mikelax (approved):
@nazar  looks good! :shipit: 
one question, have we kind of settled on dash separated class names for objects.

nazar:
>one question, have we kind of settled on dash separated class names for objects

Good question :D - CSS is hyphen based (i.e. properties) and I _think_ the general consensus is to also utilise dash separated lower case names for class names.

PR #165, opened 2018-06-29
WIP 144 Verify email API.

### LINKED ISSUE #144: Implement Resend Verification Email on Settings Page
Overview
===
On the Settings page, if the user's email is not verified, there is a button to resend the verification email. The button needs to be wired up, so when pressed it will call the appropriate Auth0 API to resend the verification email. 

Requirements
===
The API to call is: POST [/api/v2/jobs/verification-email](https://auth0.com/docs/api/management/v2#!/Jobs/post_verification_email)

The schema for the body for the POST looks to be of the form (`client_id` can be left empty): 
```json
{
  "user_id": "google-oauth2|1234",
  "client_id": ""
}
```

mikelax:
Closes #144

### DIFF
--- a/client/src/components/Settings/index.js
+++ b/client/src/components/Settings/index.js
@@ -64,9 +64,8 @@ class Settings extends Component {
         <div className="Settings">
           <Container>
             {
-              displaySuccess ? (
+              displaySuccess &&
                 <SuccessToast text='Your changes have been saved, now back to the adventure!' />
-              ) : null
             }
 
             <Grid>
@@ -158,15 +157,11 @@ class Settings extends Component {
 
   _emailVerification = () => {
     const emailVerified = _.get(this.state, 'profile.email_verified');
-    const hasEmailVerifiedKey = _.chain(this.state).get('profile').keys().includes('email_verified').value();
+    const displayVerifyEmailButton = _.has(this.state, 'profile.email_verified');
 
-    if (hasEmailVerifiedKey) {
-      return !emailVerified
-        ? (<Button>Resend Verification Email</Button>)
-        : (<Label pointing='left'>Verified <Icon color='green' size='big' name='check circle' /></Label>)
-    } else {
-      return null;
-    }
+    return displayVerifyEmailButton && !emailVerified ?
+      <Button onClick={this._handleResendEmailVerification}>Resend Verification Email</Button> :
+      <Label pointing='left'>Verified <Icon color='green' size='big' name='check circle' /></Label>;
   };
 
   _handleImage = (e) => {
@@ -190,7 +185,8 @@ class Settings extends Component {
 
   _handleResendEmailVerification = () => {
     return axiosApi
-      .post('/verify-email');
+      .post('/verify-email')
+      .then(() => this.setState({ displaySuccess: true }));
   };
 
   _valid = () => {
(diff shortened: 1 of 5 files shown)


client/src/components/Settings/index.js
```diff
@@ -158,6 +156,19 @@ class Settings extends Component {
     );
   };
 
+  _emailVerification = () => {
+    const emailVerified = _.get(this.state, 'profile.email_verified');
+    const hasEmailVerifiedKey = _.chain(this.state).get('profile').keys().includes('email_verified').value();
```
mikelax: @nazar  i was just working on the UI update as well. Couldn't this be simplified to:  
`const displayVerifyEmailButton = _.has(this.state, 'profile.email_verified');`

client/src/components/Settings/index.js
```diff
@@ -158,6 +156,19 @@ class Settings extends Component {
...
+
+    if (hasEmailVerifiedKey) {
```
mikelax: @nazar  for this block I used the following? ie. for now if we can't verify an email just show it as verified. 

```
                      {
                        displayVerifyEmailButton && !emailVerified ?
                          <Button>Resend Verification Email</Button> :
                          (<Label pointing='left'>Verified <Icon color='green' size='big' name='check circle' /></Label>)
                      }
```

client/src/components/Settings/index.js
nazar: I think the `email_verified` is absent in accounts that do not include email verification (i.e. twitch).

But other accounts, like google auth, do include the `email_verified`. Hence I'm checking for the presence of the key rather than it's value as `_.get(this.state, 'profile.email_verified')` when false will be equivalent to the key being missing

client/src/components/Settings/index.js
nazar: sure - go for it 👍

client/src/components/Settings/index.js
mikelax: @nazar  sorry I meant just the `hasEmailVerifiedKey` line. ie it would be: 
```
    const emailVerified = _.get(this.state, 'profile.email_verified');
    const displayVerifyEmailButton = _.has(this.state, 'profile.email_verified');
```

mikelax:
@nazar  this should be good now. I have checked with states for user/pass where the email was verified and not verified and with twitch accounts. 

As a separate PR we need to clean up the `SuccessToast` component to remove `componentWillReceiveProps` and a general refactoring.

client/src/components/Settings/index.js
```diff
@@ -158,15 +157,11 @@ class Settings extends Component {
...
-    const hasEmailVerifiedKey = _.chain(this.state).get('profile').keys().includes('email_verified').value();
-
-    if (hasEmailVerifiedKey) {
-      return !emailVerified
-        ? (<Button>Resend Verification Email</Button>)
-        : (<Label pointing='left'>Verified <Icon color='green' size='big' name='check circle' /></Label>)
-    } else {
-      return null;
-    }
+    const displayVerifyEmailButton = _.has(this.state, 'profile.email_verified');
```
nazar: NICE!!!1

nazar (approved):
🚀 👨‍🚀

Issue #163, opened 2018-06-29
Quick  character edit on game messages page


mikelax:
Overview
===
This ticket will cover the use case to allow the user or GM to edit a character and update a small subset of fields. 

Requirements
===
* A player should be able to quick edit their own character.
* A GM can quick edit any accepted character that is posting in messages page. 
* The edit UI will be activated from a menu by hover / click on character profile image in messages page. 
* The options in the menu (for now) will be:  Quick Update
* This menu will be different from the GM options menu, which will contain GM specific actions and features. 
* The "quick edit" should also have a link to (Edit) the Full Character
* To start, the only field that we should expose as editable is current hit points. We should display he max hit points, but this is read only and for reference. Ie:
  * ____ / 12 hit points
  * We should also display, read only for now, the hit die. Ie: 1d8

PR #169, opened 2018-07-02
Quick edit game player character

nazar:
References #163

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -50,6 +50,7 @@
     "source-map-support": "0.5.4",
     "sql-fixtures": "1.0.0",
     "subscriptions-transport-ws": "0.9.7",
+    "tsml": "^1.0.1",
     "winston": "2.4.1",
     "yup": "0.24.1"
   },
--- a/api/src/services/characters/quickUpdateCharacterAndAddUpdateGameMessage.js
+++ b/api/src/services/characters/quickUpdateCharacterAndAddUpdateGameMessage.js
@@ -1,6 +1,7 @@
 import _ from 'lodash';
 import Bluebird from 'bluebird';
 import { transaction as knexTransaction } from 'objection';
+import tsml from 'tsml';
 
 import Character from 'models/character';
 
@@ -37,7 +38,7 @@ export default function ({
   function doUpdateCharacter() {
     const cleanInput = _(input)
       .chain()
-      .omit(['changeDescription', 'changeMeta'])
+      .omit(['changeMeta'])
       .value();
 
     return updateCharacter({
@@ -52,13 +53,13 @@ export default function ({
     const changeMessageContent = _(changeMeta)
       .chain()
       .map((change) => {
-        return (`
-          <div class="change-meta">
+        return (tsml`
+          <p class="change-meta">
             <strong>${changeDescription || 'Changed'}</strong>: 
             changed ${change.attributeDescription} 
             from <span class="before">${change.oldValue} 
             to <span class="after">${change.newValue}</span></span>    
-          </div> 
+          </p> 
         `);
       })
       .join()
--- a/api/src/services/characters/quickUpdateCharacterAndAddUpdateGameMessage.js
+++ b/api/src/services/characters/quickUpdateCharacterAndAddUpdateGameMessage.js
@@ -72,7 +72,7 @@ export default function ({
         user,
         input: {
           gameId,
-          postType: 'ic',
+          postType: 'icm',
           characterId: character.id,
           message: changeMessageContent
         },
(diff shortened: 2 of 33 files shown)


nazar:
I'm going to check out React Semantic docs to see if it supports the new 16.3 Context in the Popup component as I'm not a fan in how I'm passing the `gameId` down the the modal.

nazar:
It doesn't look like [the code for React 16.3](https://github.com/Semantic-Org/Semantic-UI-React/pull/2880) has been released yet.

I'll keep an 👁 on semantic and update once React 16.3 changes are released.

api/src/services/characters/quickUpdateCharacterAndAddUpdateGameMessage.js
```diff
@@ -0,0 +1,91 @@
...
+
+import Character from 'models/character';
+
+import CreateMessage from 'services/gameMessages/createGameMessage';
+import updateCharacter from 'services/characters/updateCharacter';
+
+import serviceExecutor from 'utils/serviceExecutor';
+
+
+export default function ({
```
mikelax, line 14: @nazar  should this be a named function? is there an eslint rule for this?

mikelax (commented):
left one small comment. 
Also, where is the logic that looks at the type of game to decide what fields should be included in the quick edit modal?

api/src/services/characters/quickUpdateCharacterAndAddUpdateGameMessage.js
nazar: A function name is optional when it is exported as the default function.

api/src/services/characters/quickUpdateCharacterAndAddUpdateGameMessage.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  function generateGameMessage() {
+    const { changeMeta, changeDescription } = input;
+
+    const changeMessageContent = _(changeMeta)
+      .chain()
+      .map((change) => {
+        return (`
+          <div class="change-meta">
```
mikelax: @nazar  should this be a `p` instead of `div` to match the messages generated by the editor?
Also, when I look at the html that was saved to the game_messages table, I don't see the class attribute, is it possible that is getting stripped somewhere?

api/src/services/characters/quickUpdateCharacterAndAddUpdateGameMessage.js
```diff
@@ -0,0 +1,91 @@
...
+            <strong>${changeDescription || 'Changed'}</strong>:
```
mikelax: @nazar  should the value here `changeDescription` be stored in the character_logs table in the change_description field? I think it should, right? 

Eventually we are going to allow viewing the "history", or changelog of a Character, and it would be nice to see the the human entered value for the description rather than the generic `Character stats updated`.

api/src/services/characters/quickUpdateCharacterAndAddUpdateGameMessage.js
```diff
@@ -0,0 +1,91 @@
...
+  function startTransaction() {
+    return knexTransaction
+      .start(Character.knex())
+      .then(t => (transaction = t));
+  }
+
+  function doUpdateCharacter() {
+    const cleanInput = _(input)
+      .chain()
+      .omit(['changeDescription', 'changeMeta'])
```
mikelax: @nazar should `changeDescription` be removed from this line here so the user entered description will override the default `Character stats updated` message?

Update:  I tried removing the field but then the save function doesn't work, so I must be missing the reason for it.

api/src/services/characters/quickUpdateCharacterAndAddUpdateGameMessage.js
nazar: >should the value here changeDescription be stored in the character_logs table in the change_description field? I think it should, right?

`changeDescription` should be stored `character_logs` - I'll double check 👍

nazar: Nice catch @mikelax

api/package.json
```diff
@@ -50,6 +50,7 @@
     "source-map-support": "0.5.4",
     "sql-fixtures": "1.0.0",
     "subscriptions-transport-ws": "0.9.7",
+    "tsml": "^1.0.1",
```
mikelax, line 53: @nazar  I'll merge this PR, but do you think this lib is necessary? I saw its not maintained anymore, though I assume the function it performs is really straight forward so no need for updates??

nazar: The lib is nice in that it strips whitespace and newlines added in backtick quoted multiline strings.

The [code is < 10 lines](https://github.com/rvagg/archived-tsml/blob/master/tsml.js) - I did attempt to copypasta it but eslint disliked the non fancy es6 formats and I was too short on time to es5 -> es6 it.

Issue #167, opened 2018-06-29
Refactor legacy React lifecycle events


mikelax:
Overview
===
We are down to only a handful of files that are still using legacy lifecycle events in the client app. We should refactor them to remove all the warnings. 

Requirements
Lint results from client app. 
I spent a few minutes looking at the `SuccessToast.js` component and was trying to figure out the best refactor for this class. Currently I think there is a bug (or at least memory leak). To get the warning in the browser, load the Settings page, then in the react dev console for the `Settings` component, toggle the `displaySuccess` attribute . false -> true -> false -> true -> false
You should see the warning about calling setState for something that isn't loaded on the DOM. 

```
12:50 $ yarn lint
yarn run v1.7.0
$ eslint .

/Users/mike/git/mikelax/forgingadventures/client/src/components/App/App.js
  26:1  warning  componentWillMount is deprecated since React 16.3.0, use UNSAFE_componentWillMount instead, see https://reactjs.org/docs/react-component.html#unsafe_componentwillmount  react/no-deprecated

/Users/mike/git/mikelax/forgingadventures/client/src/components/Games/components/GameDetailsForm/index.js
  16:1  warning  componentWillMount is deprecated since React 16.3.0, use UNSAFE_componentWillMount instead, see https://reactjs.org/docs/react-component.html#unsafe_componentwillmount                       react/no-deprecated
  16:1  warning  componentWillReceiveProps is deprecated since React 16.3.0, use UNSAFE_componentWillReceiveProps instead, see https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops  react/no-deprecated

/Users/mike/git/mikelax/forgingadventures/client/src/components/Games/components/GameSearch/index.js
  11:1  warning  componentWillMount is deprecated since React 16.3.0, use UNSAFE_componentWillMount instead, see https://reactjs.org/docs/react-component.html#unsafe_componentwillmount                       react/no-deprecated
  11:1  warning  componentWillReceiveProps is deprecated since React 16.3.0, use UNSAFE_componentWillReceiveProps instead, see https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops  react/no-deprecated

/Users/mike/git/mikelax/forgingadventures/client/src/components/Games/View/components/CreateLoungeMessage.js
[...]

PR #173, opened 2018-07-06
update lifecycle methods

nazar:
References #167


client/src/components/shared/RichEditor/index.js
```diff
@@ -25,16 +25,18 @@ export default class RichTextDisplay extends Component {
...
-  componentWillMount() {
+  componentDidMount() {
     const { message } = this.props;
 
     this.setState({ message });
   }
 
-  componentWillReceiveProps(nextProp) {
-    const { message } = nextProp;
+  componentDidUpdate() {
```
mikelax, line 34: @nazar  question, how come you used this method vs. `getDerivedStateFromProps` in this case?

same question for the SuccessToast component.

mikelax (approved):
🚀 
one small question, otherwise :shipit:

client/src/components/shared/RichEditor/index.js
nazar: `getDerivedStateFromProps` is triggered on every single render cycle and It is meant to be used in [rare cases](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#when-to-use-derived-state). `componentDidUpdate` is triggered when either `props` or `state` change.

You can try substituting `getDerivedStateFromProps` and test to see if it is an improvement.

PR #174, opened 2018-07-06
172 Update semantic UI and css.

### LINKED ISSUE #172: Update React Semantic library, which just released version 0.82, the first version aimed at React 16 and compatible with the new React Context API.
Both the React and CSS Semantic packages need to be updated.

mikelax:
Closes #172


nazar (approved):
💯 :shipit: 🚀

Issue #177, opened 2018-07-11
Implement User "Read Indicator" for Game Messages


mikelax:
Overview
===
As a user participating in a Game, when viewing Game Messages, I want an option to click to indicate I have read up to a certain point in the Messages Thread. The idea is basically to implement a bookmark that the User can manually move as they read through the Messages. 

The "Bookmarks" will be viewable to the other Players and GM in the Game, but not public Users viewing the Game. 

Requirements
===
The data to track how far along a player is within a game can be stored as a new column on the `game_players` table. 

### Simple Mock up
![progress_indicator](https://user-images.githubusercontent.com/2080701/43092356-f75f95b6-8e7a-11e8-9a34-61e206c0ebfa.png)

nazar:
Hey @mikelax 

I need a second opinion on naming (names r so hard) the table, considering that we also plan to add a second "bookmark" feature into game messages for players to bookmark any message.

Table candidate names:

* game_message_read_bookmarks
* game_message_read_progress
* game_message_read_indicators
* game_message_progress
* mike_can_think_of_a_good_table_name

mikelax:
@nazar maybe we should stay away from the term `bookmark` for this feature, as it will then be confusing when we introduce actual bookmarks. 

how about a table name of:  `game_players_progress`?
Also, as an option, do you think we should store the attributes in the `game_players` table? Or is that overloading the use of that table?

PR #179, opened 2018-07-11
WIP 178 hide quote button for non-player.

mikelax:
This was a start to hide the Quote buttons for non-players. 

@nazar  I have a question on this. Currently we are only looking at if the user is logged in or not in the `_viewingControls` function where we set the `canPost` attribute. 

This also needs to check if the user is a player in the game. 
In `ViewGameMessages`, I was going to pass a new property of `myGamePlayer={myGamePlayer}` to `GameMessages`. I then realized though that this prop would then have to be passed down to: 

MessagesRenderer -> GameMessageContainer -> GameMessage

I am not sure if this is the "right" approach, plus I wasn't sure if that was the right approach given the props we may need to track for #177

### DIFF
--- a/client/src/components/Games/View/ViewGameLoungeMessages.js
+++ b/client/src/components/Games/View/ViewGameLoungeMessages.js
@@ -1,7 +1,9 @@
+import PropTypes from 'prop-types';
 import React from 'react';
-import { Segment } from 'semantic-ui-react';
+import { Message, Segment } from 'semantic-ui-react';
 import { graphql } from 'react-apollo/index';
-import { compose, pure } from 'recompose';
+import { compose } from 'recompose';
+import { connect } from 'react-redux';
 
 import GameHeader from './GameHeader';
 import CreateLoungeMessage from './components/CreateLoungeMessage';
@@ -10,7 +12,7 @@ import GameLoungeMessages from './components/GameLoungeMessages';
 import { gameQuery } from '../queries';
 
 function GameLoungeMessagesView (props) {
-  const { match: { params: { id } }, data: { game } } = props;
+  const { match: { params: { id } }, data: { game }, authorisation: { isAuthenticated } } = props;
 
   return (
     <React.Fragment>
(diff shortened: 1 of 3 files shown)


client/src/components/Games/View/ViewGameMessages.js
```diff
@@ -20,7 +20,7 @@ function ViewGameMessages(props) {
     <React.Fragment>
       <GameHeader game={game} />
 
-      <GamesMessages gameId={game.id}/>
+      <GameMessages gameId={_.toInteger(game.id)}/>
```
mikelax: @nazar  is the `toInteger` correct, or should the "id" attribute already be a number within the game object?

nazar: `game.id` is defined in the graphQL as an `ID` scalar type, which a string.

mikelax:
thanks for the work @nazar . I was going to ask about changing the proptype to `string` i guess that is what we should do for props that are scalars. 

With regards to the other refactor, I see what you did, would another option have been to use the new context API? It must have been late as I didn't think of that last night, until this morning while walking the dog.

client/src/components/Games/View/ViewGameLoungeMessages.js
```diff
@@ -19,15 +21,36 @@ function GameLoungeMessagesView (props) {
       <GameLoungeMessages gameId={id}/>
 
       <Segment>
-        <CreateLoungeMessage gameId={id}/>
+        { isAuthenticated && <CreateLoungeMessage gameId={id}/> }
```
mikelax, line 24: @nazar  what do you think about this pattern of using the `isAuthenticated` variable for determining the component to display? 

other option would be to do an if check in the render function before the return and assign the correct component to a variable.

nazar: >what do you think about this pattern of using the isAuthenticated variable for determining the component to display?

It's perfect 👍 💯

nazar (approved):
🚀

Issue #168, opened 2018-07-01
Add Skills to Characters for 5e engine


mikelax:
Overview
===
There is a standard set of Skills that are used in 5e. We should add this list to the Character Settings page. 

Requirements
===
Here is the standard list of skills. NOTE, *maybe* at some point in the future we would allow users to create custom skills, but not for v1. 
```
Acrobatics(Dex)      
Animal Handling (Wis)
Arcana (Int)         
Athletics (Str)      
Deception (Cha)      
History (Int)        
Insight (Wis)        
Intimidation (Cha)   
Investigation (Int)  
Medicine (Wis)       
Nature (Int)         
Perception (Wis)     
Performance (Cha)    
Persuasion (Cha)     
Religion (Int)       
Sleight of Hand (Dex)
Stealth (Dex)        
Survival (Wis)       
```

Each Skill is associated with one of the main Abilities, which will influence the Skill Bonus. 
On the Create/Edit Character page, we should display the list of Skills in a new section. The section should be displayed below the Abilities section. 
For each row in the skills section, it should display the Skill Name, related ability, checkbox to indicate proficiency, and a textbox for "Bonus". 
The Bonus textbox will default to display a value of the related Ability Modifier. 
The user will be able to select one or more skills they are proficient in, if the checkbox is checked the Character's Proficient Bonus should be added to the value in the textfield. 
For v1, the values of the Skills Bonus will be fully editable by the User. 

Data Storage
---
For data storage within the validator, database, etc. we should store the following for each skill. 
* skillName
* ability
* proficient (bool)
* bonus (int)

Examples
---
Here are a couple of examples of newer character sheets. 
https://www.dndbeyond.com/profile/mikelax/characters/548273
https://mobile.twitter.com/mattcolville/status/1012911296784654336/photo/1
DND Beyond just [revamped their character sheet](https://www.dndbeyond.com/posts/262-the-character-sheet-revamp-is-live-now) features.

Bonus
---
Some classes allow for a trait called `Expertise`. So at a certain level, maybe 3, the character would be able to double the proficiency bonus for one or more of their Skills. Should we just allow this to be "manual" for now and left up to the user (ie. they can just update the bonus value).

PR #182, opened 2018-07-16
Add Skills to Characters for 5e engine

nazar:
References #168


mikelax:
@nazar  this is looking really nice! I have one small comment on the UI for how each Skill component should be rendered. I think it should be something like the following. 
The textfield can be made much smaller to create space as it will only hold a single digit. 
The change is to make it obvious to the user that the toggle element applies to Proficiency. 

SkillName (Ability)     Proficient (toggle element)   textfield
ex: 
Acrobatics (Dex)    Proficient (toggle)   textfield


Also, it appears that all the elements are top-aligned within the Skill Component. Can we make this middle aligned?

mikelax (approved):
🚀

PR #185, opened 2018-07-24
Implement User "Read Indicator" for Game Messages

nazar:
Refs #177


api/src/schemas/gamePlayer.js
```diff
@@ -83,11 +87,12 @@ export const gamePlayerResolvers = {
       schemaScopeGate(['create:games'], context, () => {
         return getOrCreateUserByAuth0Id(context.req.user.sub)
           .then((user) => {
-            input.userId = user.id;
```
mikelax, line 86: @nazar  just a style reason for moving this line into the `.insert` function?

client/src/components/Games/View/components/GameMessages/index.js
```diff
@@ -338,14 +347,26 @@ class GameMessage extends Component {
   };
 
   _viewingControls = () => {
-    const { isCurrentUserMessage, isMemberOfGame } = this.props;
+    const { isCurrentUserMessage, isMemberOfGame, gameMessage } = this.props;
     const canEdit = isCurrentUserMessage && this._editableMessage();
+    const canBookmark = !(isCurrentUserMessage) && this._editableMessage();
```
mikelax: @nazar  maybe rename this variable so we don't get confused with actual bookmark feature? 
maybe `canUpdateProgress`?

Also, I don't think that we should hide the feature for your own messages.

api/src/schemas/gamePlayer.js
nazar: yep - trying to avoid mutable data and also being fan of `{...object, a: b}` immutable variation.

mikelax (approved):

mikelax (approved):

PR #186, opened 2018-07-30
closes #69 - Display issue on games list page, newly created game is …

### LINKED ISSUE #69: Display issue on games list page, newly created game is added twice
Overview
===
After creating a new game on the /games/create page, as a user I am redirected back to the /games list page. The newly created game is displayed in the list twice. We have seen this bug present itself if this game was the first game ever created, or if there were already two games created. 

@nazar believes the issue may be related to infinite load. 
>I have a feeling it is an issue with the infinite load - i.e the page view and the inf load both trigger at the same time

screen shot 2018-02-08 at 8 33 28 am

nazar:
…added twice

### DIFF
--- a/client/src/components/Games/List/index.js
+++ b/client/src/components/Games/List/index.js
@@ -79,7 +79,9 @@ export default connect(
 
 class ListGamesPure extends Component {
 
-  canPage = true;
+  state = {
+    canPage: true
+  };
 
   render() {
     const { data: { games, loading } } = this.props;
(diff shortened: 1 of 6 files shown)


client/src/components/Games/List/index.js
```diff
@@ -100,14 +102,15 @@ class ListGamesPure extends Component {
...
+    const { fetchMore, data: { games, loading } } = this.props;
+    const { canPage } = this.state;
     const gamesCount = _.get(games, 'length');
 
-    if (this.canPage && inView && !(loading) && gamesCount) {
+    if (canPage && inView && !(loading) && gamesCount) {
       fetchMore()
         .then(({ data: { games } }) => {
           if (_.isEmpty(games)) {
-            this.canPage = false;
```
mikelax, line 110: 😮

nazar: images

mikelax (approved):

PR #190, opened 2018-09-09
refactor: Update code style for react classes, function components.

mikelax:


client/src/components/Characters/index.js
```diff
@@ -7,20 +7,15 @@ import EditCharacter from './Edit';
...
-          <Switch>
-            <Route exact path={`${match.url}/create`} component={AuthGuard(CreateCharacter)}/>
-            <Route exact path={`${match.url}/:id/edit`} component={AuthGuard(EditCharacter)}/>
-          </Switch>
-        </Container>
-      </div>
-    );
-  }
-}
+export default Characters;
```
nazar, line 21: Hey @mikelax 

The `const` and `export` can both be on one line if we use the `function` form:

```javascript
export default function Characters({ match }) {
...
}
```

client/src/components/Games/View/index.js
```diff
@@ -12,22 +12,16 @@ import AuthGuard from 'components/shared/AuthGuard';
...
+      <Route path={`${match.url}:id/edit`} component={AuthGuard(EditGame)}/>
+      <Route path={`${match.url}:id/lounge`} component={ViewGameLoungeMessages} />
+      <Route path={`${match.url}:id/messages`} component={ViewGameMessages} />
+      <Route path={`${match.url}:id/join`} component={AuthGuard(JoinGame)}/>
+      <Route path={`${match.url}:id`} component={ViewGameDetails} />
+    </Switch>
+  </div>
+);
+
+export default Game;
```
nazar, line 27: Hey @mikelax 

The `const` and `export` can both be on one line if we use the `function` form:

```javascript
export default function Game({ match }) {
...
}
```

client/src/components/Games/index.js
```diff
@@ -8,21 +8,16 @@ import ViewGame from './View';
...
-            <Route path={`${match.url}/create`} component={AuthGuard(CreateGame)}/>
-            <Route strict path={`${match.url}/`} component={ViewGame}/>
-            <Route strict exact path={`${match.url}`} component={ListGames}/>
-          </Switch>
-        </Container>
-      </div>
-    );
-  }
-}
+export default Games;
```
nazar, line 23: See my previous comments.

client/src/components/Login/index.js
```diff
@@ -7,19 +7,14 @@ import AlmostFinished from './AlmostFinished';
...
-        <Switch>
-          <Route path={`${match.url}/callback`} component={Callback}/>
-          <Route path={`${match.url}/almost-finished`} component={AuthGuard(AlmostFinished)}/>
-          <Route exact path={`${match.url}/`} component={Login}/>
-        </Switch>
-      </div>
-    );
-  }
-}
+export default LoginView;
```
nazar, line 20: See my previous comments

nazar (changes requested):
Thanks for the refactor @mikelax; it looks great 👍. I left a few comments re default function defs.

client/src/components/Characters/index.js
mikelax: Hey @nazar. After I looked at some of the functions in DS repo I realized we did the export the other way. I *thought* you might call this out. I’ll change when I get to office.

nazar: Hey @mikelax! On second thought, I think the way you have is better as the `return` statement is not required. 

It's funny that with `const` we don't need the extra `return (...)` but need the `export` on a seperate line. And vice-versa with the `function` form....

nazar (approved):

PR #192, opened 2018-11-03
chore: get build working for client.

mikelax:
This is a WIP. 
It's been a while since looking at the fa client, but I could not get a `docker build` or even with `yarn build` to work unless I made this change in the webpack config to comment out the eslint plugin. 
You'll notice I even needed to make a change to the Dockerfile to get it to build???

I tried comparing to the DS client project and webbpack configs and I can't notice any differences. 
Any ideas? 

The goal for this change is to explore taking the client build and publishing it to netlify to eliminate all the docker resources, config, etc for the client app. Should simplify (and reduce cost) a bunch.


client/Dockerfile
```diff
@@ -13,9 +13,8 @@ COPY /config /usr/src/app/config
 COPY /public /usr/src/app/public
 COPY /scripts /usr/src/app/scripts
 COPY /src /usr/src/app/src
-
-# copy specific build target env files
-COPY .env.* .
+COPY .env.* /usr/src/app/
+COPY .eslint* /usr/src/app/
```
nazar, line 18: I'm surprised this change was required since line 8 already defines the working directory `WORKDIR /usr/src/app`, so a `COPY .env.* .` should I imagine be the same as `.env.* /usr/src/app/`

client/config/webpack.config.prod.js
```diff
@@ -330,26 +327,26 @@ module.exports = {
...
+    //   },
+    //   output: {
+    //     comments: false,
+    //     // Turned on because emoji and regex is not minified properly using default
+    //     // https://github.com/facebookincubator/create-react-app/issues/2488
+    //     ascii_only: true,
+    //   },
+    //   sourceMap: shouldUseSourceMap,
+    // }),
     // Note: this won't work without ExtractTextPlugin.extract(..) in `loaders`.
```
nazar: oh wow! I'm surprised this stopped working. I'm sure we use this in match as well? 🤔

client/Dockerfile
mikelax: @nazar  I was getting a docker error that when copying multiple files, the destination must be a directory instead of a single file. ie. `/usr/src/app/` instead of `.`, maybe this would have also worked but I didn't try it `./`

client/config/webpack.config.prod.js
mikelax: we do use it in Match and the config is the same. I double checked and in Match the dockerfile is doing a `yarn build` which is what I am doing here. 🤷‍♂️

nazar: yeah - that is defo super weird. I'm check what happens on my local dev and report back.

mikelax:
@nazar  still a WIP for netlify but I have a [preview site working](https://5bde254482d3f1047bb0888b--playforgingadventures.netlify.com/), obviously not hooked up to an api yet.

mikelax:
@nazar  I took another look at the webpack prod configs between DS and FA, and I noticed one difference is with the stylus loaders. In DS we have a single `test` command, and also one uses rupture vs. jeet. I tried to update FA to use a single loader with `rupture` but I was getting a different error, I think before it even got to the uglifyjs portion. I know DS doesn't use semantic ui, but I 🤷‍♂️ 

```
Failed to compile.

./node_modules/semantic-ui-css/semantic.min.css
Module build failed: ModuleBuildError: Module build failed: ParseError: /Users/mike/git/mikelax/forgingadventures/client/node_modules/semantic-ui-css/semantic.min.css:27:4940
```

client/config/webpack.config.prod.js
nazar: Digging :shovel: deeper, this appears to be a CRA 1.x issue (solved in 2.x) related to the useage of the `query-string` library [here](https://github.com/mikelax/forgingadventures/blob/master/client/src/components/shared/UberPaginator.js#L7) with more info in [this CRA thread](https://github.com/facebook/create-react-app/issues/3734).

The prod builds fine with the Uglifier comments reverted.

Looks like we'll have to swap this out with the `qs` library, which we use on DS.

nazar:
>Failed to compile.
>
>./node_modules/semantic-ui-css/semantic.min.css
Module build failed: ModuleBuildError: Module build failed: ParseError: /Users/mike/git/mikelax/forgingadventures/client/node_modules/semantic-ui-css/semantic.min.css:27:4940

Hmmm... I'm not seeing this on my newly built dev containers. Is this from `yarn build` inside a container or on the host?

